PR #4, opened 2019-05-24
CLIENT-4862 | Convert RTCMonitor to TypeScript

charliesantos:
**Contributing to Twilio**

### JIRA link(s):

- [CLIENT-4862](https://issues.corp.twilio.com/browse/CLIENT-4862)

### Description

* Convert rtc/monitor.js into typescript
* Add unit tests

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


ryan-rowland:
@charliesantos The docs are currently needed for publishing thee `gh-pages` branch. Would be interest to find a better way to do this and automate this.

lib/twilio/connection.ts
```diff
@@ -7,23 +7,19 @@ import { EventEmitter } from 'events';
 import Device from './device';
 import DialtonePlayer from './dialtonePlayer';
 import { Region } from './regions';
+import RTCMonitor from './rtc/monitor';
```
ryan-rowland: How about `import RTCMonitor, { RTCSample } from './rtc/monitor'?

charliesantos:
> @charliesantos The docs are currently needed for publishing thee `gh-pages` branch. Would be interest to find a better way to do this and automate this.

Ok we can talk about it as a separate discussion. I will put them back and won't do the changes in this PR.

lib/twilio/connection.ts
charliesantos: `RTCSample` is inside `RTCMonitor` namespace and cannot be exported in this manner. We can do it if we don't put `RTCSample` inside a namespace.

We have code that uses this pattern (namespacing). What is our convention/rule regarding namespacing? My suggestion is to remove namespaces so it is more aligned with ES Modules, where the whole file is essentially the 'namespace', plus, we can easily `import` using the way you suggested above :). Let me know what you think.

ryan-rowland: I think we need to remove this from the RTCMonitor namespace, otherwise we need to expose RTCMonitor to the customer, and it's not part of the public API. Let's define it in its own file and pull it in to both `monitor.ts` and `connection.ts`. WDYT?

I've been using namespacing for stuff that is directly tied to a class, and would otherwise be repetitive. If `RTCSample` were in the `RTCMonitor` namespace I'd probably want to call it `RTCMonitor.Sample` rather than `RTCMonitor.RTCSample` just due to redundancy.

lib/twilio/rtc/monitor.ts
```diff
@@ -0,0 +1,593 @@
...
+
+  /**
+   * The PeerConnection to monitor.
+   */
+  private _peerConnection: IPeerConnection;
+
+  /**
+   * Method to get stats from a PeerConnection object. Overrides getRTCStats library
+   */
+  private _getRTCStats: (peerConnection: IPeerConnection) => IRTCStats;
```
ryan-rowland: Are you running tslint? This should be throwing errors since it's not in alphabetical order.

lib/twilio/rtc/monitor.ts
```diff
@@ -0,0 +1,593 @@
...
+        packetsSent: stats.packetsSent,
+      },
+    };
+  }
+
+  /**
+   * Get stats from the PeerConnection and add it to our list of samples.
+   * @returns A universally-formatted version of RTC stats.
+   */
+  private _fetchSample(): Promise<RTCMonitor.RTCSample | void> {
```
ryan-rowland: Hmm, this shouldn't be void. We should probably let the error pass through, and catch it on Connection.

lib/twilio/connection.ts
charliesantos: Defining it in its own file sounds better. I'll put it as `rtc/sample.ts`

lib/twilio/rtc/monitor.ts
charliesantos: Yes I'm running tslint. Not sure why it did not catch it. I will move this so it's in alphabetical order.

lib/twilio/rtc/monitor.ts
charliesantos: Actually, this doesn't need to return anything. It's within the setInterval above (for on 'sample'), And this function is already emitting both 'sample' and 'error' events

charliesantos:
@ryan-rowland I finished merging from master and addressed review comments.

lib/twilio/rtc/monitor.ts
```diff
@@ -0,0 +1,491 @@
...
+      this.enable(this._peerConnection);
+    }
+  }
+
+  /**
+   * Start sampling RTC statistics for this {@link RTCMonitor}.
+   * @param peerConnection - A PeerConnection to monitor.
+   * @returns The current {@link RTCMonitor}.
+   */
+  enable(peerConnection: IPeerConnection): RTCMonitor {
```
ryan-rowland: I believe we can use `this` instead of `RTCMonitor` here to make it clear it returns itself.

PR #5, opened 2019-05-30
CLIENT-6145 CLIENT-5873 | Restart ICE when PeerConnection is disconnected

charliesantos:
**Contributing to Twilio**

### JIRA link(s):

- [CLIENT-6145](https://issues.corp.twilio.com/browse/CLIENT-6145)
- [CLIENT-5873](https://issues.corp.twilio.com/browse/CLIENT-5873)

### Description

ICE Connections will now attempt to reconnect when they transition to the `failed` state, rather than immediately disconnecting. See [Jira CLIENT-6145](https://issues.corp.twilio.com/browse/CLIENT-6145) for more details.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/lib/twilio/connection.ts
+++ b/lib/twilio/connection.ts
@@ -45,12 +45,15 @@ const DTMF_INTER_TONE_GAP: number = 70;
 const DTMF_PAUSE_DURATION: number = 500;
 const DTMF_TONE_DURATION: number = 160;
 
+const ICE_RESTART_INTERVAL = 3000;
 const METRICS_BATCH_SIZE: number = 10;
 const METRICS_DELAY: number = 20000;
 
 const WARNING_NAMES: Record<string, string> = {
   audioInputLevel: 'audio-input-level',
   audioOutputLevel: 'audio-output-level',
+  bytesReceived: 'bytes-received',
+  bytesSent: 'bytes-sent',
   jitter: 'jitter',
   mos: 'mos',
   packetsLostFraction: 'packet-loss',
@@ -123,6 +126,11 @@ class Connection extends EventEmitter {
    * will copy whatever we get from RTC stats.
    */
   private _codec: string;
+
+  /**
+   * Interval id for ICE restart loop
+   */
+  private _iceRestartIntervalId: number;
 
   /**
    * The number of times input volume has been the same consecutively.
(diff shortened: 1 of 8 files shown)


lib/twilio/connection.ts
```diff
@@ -959,6 +961,20 @@ class Connection extends EventEmitter {
     }
   }
 
+  /**
+   * Monitors and restarts ICE for the current {@link Connection}
+   */
+  private _monitorIceConnection(): void {
+    const reconnectIce = () => this.mediaStream.iceRestart().then(listen, listen);
+
+    const listen = () => this._monitor.once('disconnect', () => {
```
ryan-rowland: What happens when the call is over? Is this cleaned up?

lib/twilio/rtc/monitor.js
```diff
@@ -209,6 +214,32 @@ RTCMonitor.prototype._addSample = function _addSample(sample) {
   }
 };
 
+/**
+ * Raise disconnect event base on bytes sent and received
+ * @private
+ */
+RTCMonitor.prototype._checkIceConnection = function _checkIceConnection() {
+  // How many samples we need to check within the last few seconds (ICE_DISCONNECT_THRESHOLD)
+  const samplesNeeded = Math.floor(ICE_DISCONNECT_THRESHOLD / SAMPLE_INTERVAL);
```
ryan-rowland: Is this always correct? Does floating point math ever make this like `2.9999999`?

lib/twilio/rtc/monitor.js
```diff
@@ -209,6 +214,32 @@ RTCMonitor.prototype._addSample = function _addSample(sample) {
...
+
+  // Grab the samples we need starting from most recent.
+  const sampleBytes = this._sampleBuffer.slice(this._sampleBuffer.length - samplesNeeded)
+    .map(sample => ({ sent: sample.totals.bytesSent, received: sample.totals.bytesReceived }));
+
+  // No enough samples
+  if (sampleBytes.length < samplesNeeded && this._sampleBuffer.length < SAMPLE_COUNT_METRICS) {
```
ryan-rowland: Should this happen earlier? Can we just check `this._sampleBuffer.length` before creating `sampleBytes`?

lib/twilio/rtc/monitor.js
```diff
@@ -209,6 +214,32 @@ RTCMonitor.prototype._addSample = function _addSample(sample) {
...
+    return;
+  }
+
+  const noActivity = sampleBytes.every(bytes => sampleBytes[0].sent === bytes.sent) ||
```
ryan-rowland: For readability, I'd prefer booleans to be named such that they could only be booleans. Usually that means containing: `is`, `was`, `has`, `should`, etc...

For example, `isDataPaused` might be good here

lib/twilio/rtc/monitor.js
```diff
@@ -209,6 +214,32 @@ RTCMonitor.prototype._addSample = function _addSample(sample) {
...
+    sampleBytes.every(bytes => sampleBytes[0].received === bytes.received);
+
+  if (noActivity && Date.now() - this._lastDisconnect >= ICE_DISCONNECT_THRESHOLD) {
+    this._lastDisconnect = Date.now();
+    this.emit('disconnect', this._peerConnection);
```
ryan-rowland: I don't think we need to pass the PeerConnection here do we?

Also, not sure about `disconnect`, it's a bit vague. In fact, this should probably match the other warning events emitted by RTCMonitor. We should raise a warning when we lose ICE connectivity, and raise a warning-cleared when it's resolved. WDYT?

lib/twilio/connection.ts
charliesantos: I don't think so. It's called from the constructor, same as the rest of our subscriptions. Do we have cleanup routine that I can hook up to?

lib/twilio/rtc/monitor.js
charliesantos: I'm trying to prevent floating point numbers in case we ever change SAMPLE_INTERVAL value. For example, if . SAMPLE_INTERVAL is 2000, we get 1.5 which we don't want.

lib/twilio/rtc/monitor.js
charliesantos: yes we can!

lib/twilio/rtc/monitor.js
charliesantos: Agreed. Are you thinking we can make it a new property of the sample, and set it inside the createSample method? Or just set a threshold for the existing stat?

lib/twilio/connection.ts
```diff
@@ -261,8 +269,22 @@ class Connection extends EventEmitter {
     monitor.disableWarnings();
     setTimeout(() => monitor.enableWarnings(), METRICS_DELAY);
 
-    monitor.on('warning', this._reemitWarning);
-    monitor.on('warning-cleared', this._reemitWarningCleared);
+    monitor.on('warning', (data: Record<string, any>, wasCleared?: boolean) => {
```
ryan-rowland: Let's make sure we remember to update this with the other PR since this `data` field should have a type

PR #8, opened 2019-06-04
CLIENT-6049 | Emitting average audio levels

charliesantos:
**Contributing to Twilio**

### JIRA link(s):

- [CLIENT-6049](https://issues.corp.twilio.com/browse/CLIENT-6049)

### Description

This PR reports average audio levels to insights instead of the last emitted audio level.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/lib/twilio/connection.ts
+++ b/lib/twilio/connection.ts
@@ -1089,8 +1089,8 @@ class Connection extends EventEmitter {
   private _onRTCSample = (sample: RTCSample): void => {
     const callMetrics: Connection.CallMetrics = {
       ...sample,
-      audioInputLevel: average(this._internalInputVolumes),
-      audioOutputLevel: average(this._internalOutputVolumes),
+      audioInputLevel: Math.round(average(this._internalInputVolumes)),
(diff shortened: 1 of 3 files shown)


lib/twilio/connection.ts
```diff
@@ -1087,10 +1088,17 @@ class Connection extends EventEmitter {
    * @param sample
    */
   private _onRTCSample = (sample: RTCSample): void => {
+    const extractAverage = (items: number[]) => {
```
ryan-rowland: This runs every second! We should define this function once rather than every second. I think we actually have an `average` function defined in `util.js`

ryan-rowland: I think logic here should like like:
```
_onRTCSample = () => {
  const callMetrics = {
    audioInputLevel: average(this._internalInputVolumes),
    //...
  };
  this._internalInputVolumes.splice(0, this._internalInputVolumes.length);
  //...
```

charliesantos:
comments addressed @ryan-rowland

PR #17, opened 2019-06-28
Client 6319 | Stop sending reinvites on server hangup

charliesantos:
**Contributing to Twilio**

### JIRA link(s):

- [CLIENT-6319](https://issues.corp.twilio.com/browse/CLIENT-6319)

### Description
@ryan-rowland  I found a 100% repro steps. See jira link above.

Update summary:
- iceRestart now returns a boolean flag `canRetry` so that the calling function can decide whether it should stop retrying. To make this possible, I made reinvite its own function. It wraps the websocket calls and responses into a promise to easily determine whether the reinvite for a callsid succeeded or failed. Success means we received an `answer`, failed means we received a `hangup` from the server. This helps on stopping the iceRestart loop.
- Device.destroy() now calls disconnectAll()

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/lib/twilio/connection.ts
+++ b/lib/twilio/connection.ts
@@ -127,7 +127,7 @@ class Connection extends EventEmitter {
   /**
    * Interval id for ICE restart loop
    */
-  private _iceRestartIntervalId: number;
+  private _iceRestartIntervalId: NodeJS.Timer;
 
   /**
    * The number of times input volume has been the same consecutively.
@@ -269,16 +269,25 @@ class Connection extends EventEmitter {
     monitor.on('warning', (data: RTCWarning, wasCleared?: boolean) => {
       if (data.name === 'bytesSent' || data.name === 'bytesReceived') {
         this._log.warn('ICE Connection disconnected.');
-        clearInterval(this._iceRestartIntervalId);
-        this._iceRestartIntervalId = setInterval(
-          this.mediaStream.iceRestart.bind(this.mediaStream), ICE_RESTART_INTERVAL);
+
+        // Stop existing loops if this warning is emitted multiple times
+        this._stopIceRestarts();
+
+        this._iceRestartIntervalId = setInterval(() => {
+          this.mediaStream.iceRestart().catch((canRetry: boolean) => {
+            if (!canRetry) {
+              this._log.info('Received hangup from the server. Stopping attempts to restart ICE.');
+              this._stopIceRestarts();
+            }
+          });
+        }, ICE_RESTART_INTERVAL);
       }
       this._reemitWarning(data, wasCleared);
     });
     monitor.on('warning-cleared', (data: RTCWarning) => {
       if (data.name === 'bytesSent' || data.name === 'bytesReceived') {
         this._log.info('ICE Connection reestablished.');
-        clearInterval(this._iceRestartIntervalId);
+        this._stopIceRestarts();
       }
       this._reemitWarningCleared(data);
     });
@@ -386,6 +395,7 @@ class Connection extends EventEmitter {
       }
 
       monitor.disable();
+      this._stopIceRestarts();
       this._publishMetrics();
 
       this.emit('disconnect', this);
@@ -950,6 +960,7 @@ class Connection extends EventEmitter {
    */
   private _disconnect(message?: string | null, wasRemote?: boolean): void {
     message = typeof message === 'string' ? message : null;
+    this._stopIceRestarts();
 
     if (this._status !== Connection.State.Open
         && this._status !== Connection.State.Connecting
@@ -958,7 +969,6 @@ class Connection extends EventEmitter {
     }
 
     this._log.info('Disconnecting...');
-    clearInterval(this._iceRestartIntervalId);
 
     // send pstream hangup message
     if (this.pstream !== null && this.pstream.status !== 'disconnected' && this.sendHangup) {
@@ -1206,6 +1216,13 @@ class Connection extends EventEmitter {
 
     this.parameters.CallSid = callSid;
     this.mediaStream.callSid = callSid;
+  }
+
+  /**
+   * Stop ice restart loop
+   */
+  private _stopIceRestarts(): void {
+    clearInterval(this._iceRestartIntervalId);
   }
 }
 
--- a/lib/twilio/deferred.ts
+++ b/lib/twilio/deferred.ts
@@ -1,0 +1,55 @@
+/**
+ * @module Voice
+ * @internalapi
+ */
+
+/**
+ * Deferred Promise
+ */
+export default class Deferred {
+  /**
+   * This {@link Deferred} promise
+   */
+  private readonly _promise: Promise<any>;
+
+  /**
+   * Rejects this promise
+   */
+  private _reject: (reason?: any) => void;
+
+  /**
+   * Resolves this promise
+   */
+  private _resolve: (value?: any) => void;
+
+  /**
+   * @constructor
+   */
+  constructor() {
+    this._promise = new Promise<any>((resolve, reject) => {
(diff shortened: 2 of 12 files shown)


charliesantos:
I updated and added unit tests. This is now ready for review and merged if you approve @ryan-rowland

lib/twilio/connection.ts
```diff
@@ -269,16 +269,25 @@ class Connection extends EventEmitter {
...
-        this._iceRestartIntervalId = setInterval(
-          this.mediaStream.iceRestart.bind(this.mediaStream), ICE_RESTART_INTERVAL);
+
+        // Stop existing loops if this warning is emitted multiple times
+        this._stopIceRestarts();
+
+        this._iceRestartIntervalId = setInterval(() => {
+          this.mediaStream.iceRestart().catch((canRetry: boolean) => {
+            if (!canRetry) {
+              this._log.info('Received stopIceRestarts from the server. Stopping iceRestart...');
```
ryan-rowland: The server has no concept of `stopIceRestarts`, that's a local method we're referring to. Can we fix this message to be more precise?

lib/twilio/pstream.js
```diff
@@ -129,6 +133,19 @@ PStream.prototype._handleTransportMessage = function(msg) {
   const { type, payload = {} } = JSON.parse(msg.data);
   this.gateway = payload.gateway || this.gateway;
   this.region = payload.region || this.region;
+
+  // Resolve our reinvite handler if there's any
+  const handler = this._reInviteHandlers[payload.callsid];
```
ryan-rowland: Why is this a hash by callsid? We currently aren't able to have multiple active calls on one Device, so is this necessary?

ryan-rowland: Actually, I think we should nix all of this new logic. The way this class is set up is it emits events when a message is received, and the consuming class executes the logic in a handler. We don't pass handlers in as callbacks. This is combining two paradigms that accomplish the same thing.

ryan-rowland: Nevermind, see review

lib/twilio/connection.ts
ryan-rowland: Nevermind, see review.

ryan-rowland: See review.

lib/twilio/pstream.js
```diff
@@ -129,6 +130,19 @@ PStream.prototype._handleTransportMessage = function(msg) {
   const { type, payload = {} } = JSON.parse(msg.data);
   this.gateway = payload.gateway || this.gateway;
   this.region = payload.region || this.region;
+
+  // Resolve our reinvite handler if there's any
+  const handler = this._reinviteDeferreds.get(payload.callsid);
```
ryan-rowland: This is a `deferred` now right?

lib/twilio/pstream.js
```diff
@@ -163,6 +177,21 @@ PStream.prototype.register = function(mediaCapabilities) {
...
+
+  if (deferred) {
+    promise = deferred.promise;
+  } else {
+    promise = new Promise((resolve, reject) => {
+      this._reinviteDeferreds.set(callsid, { promise, resolve, reject });
+      this._publish('reinvite', { sdp, callsid }, false);
+    });
+  }
+  return promise;
```
ryan-rowland: You can drop `let promise` and sum up lines 184...192 like this:
```
return deferred
  ? deferred.promise
  : new Promise( /* ... */ );
```

charliesantos: Not sure if that works. I need to catch the promise object before I return. See Line 188
```
promise = new Promise((resolve, reject) => {
      this._reinviteDeferreds.set(callsid, { promise, resolve, reject });
      this._publish('reinvite', { sdp, callsid }, false);
    });
```

ryan-rowland: Ah. Then:
```
const promise = deferred
  ? deferred.promise
  : new Promise( /* ... */ );

return promise;
```

charliesantos: I updated to this (see below) and it won't catch the `promise` object. Which makes sense, because the promise body gets executed first before it returns and assigned to the promise variable.
```
  const deferred = this._reinviteDeferreds.get(callsid);
  const promise = deferred ? deferred.promise : new Promise((resolve, reject) => {
    this._reinviteDeferreds.set(callsid, { promise, resolve, reject });
    this._publish('reinvite', { sdp, callsid }, false);
  });
  return promise;
```

charliesantos: I updated this function to handle scenario where sometimes the server won't send back a response. True if signaling drops.

lib/twilio/deferred.ts
```diff
@@ -0,0 +1,55 @@
...
+ * Deferred Promise
+ */
+export default class Deferred {
+  /**
+   * This {@link Deferred} promise
+   */
+  private readonly _promise: Promise<any>;
+
+  /**
+   * Rejects this promise
```
ryan-rowland: This is a handle on the Promise's `reject` method right? Let's make this a bit clearer:
```
/**
 * The Promise's reject method.
 */
```

Same for resolve

lib/twilio/deferred.ts
```diff
@@ -0,0 +1,55 @@
...
+   */
+  private _reject: (reason?: any) => void;
+
+  /**
+   * Resolves this promise
+   */
+  private _resolve: (value?: any) => void;
+
+  /**
+   * @constructor
```
ryan-rowland: This seems erroneous, does this make a difference in typedoc?

lib/twilio/deferred.ts
```diff
@@ -0,0 +1,55 @@
...
+   */
+  constructor() {
+    this._promise = new Promise<any>((resolve, reject) => {
+      this._resolve = resolve;
+      this._reject = reject;
+    });
+  }
+
+  /**
+   * @returns The {@link Deferred} promise
```
ryan-rowland: nitpick: Promise (capitalized)

lib/twilio/deferred.ts
```diff
@@ -0,0 +1,55 @@
...
+   */
+  get promise(): Promise<any> {
+    return this._promise;
+  }
+
+  /**
+   * Rejects this promise
```
ryan-rowland: This comment and below (resolve) can stay unchanged

lib/twilio/pstream.js
```diff
@@ -163,6 +178,14 @@ PStream.prototype.register = function(mediaCapabilities) {
   this._publish('register', regPayload, true);
 };
 
+PStream.prototype.reinvite = function(sdp, callsid) {
+  const deferred = this._reinviteDeferreds.get(callsid) || new Deferred();
+  this._reinviteDeferreds.set(callsid, deferred);
+
+  this._publish('reinvite', { sdp, callsid }, false);
```
ryan-rowland: Do we want to do this if we already have a deferred? Isn't the point to return `this._reinviteDeferreds.get(callsid).promise` if it already exists instead?

lib/twilio/deferred.ts
ryan-rowland: To be clear, I'm saying they are accurate since they're methods we call, and shouldn't be updated like I requested above for _reject and _resolve.

lib/twilio/pstream.js
charliesantos: Yes we want to publish to server because we don't have a guaranteed response from the server especially when the signaling is dropped. E.g. we call iceRestart 10 times when we lose connection. Only the first retry will publish and the remaining will not publish to server. Once the connection is restored with a new instance of WSTransport, we won't get a response back.

PR #18, opened 2019-07-02
CLIENT-6335 | Making sure we only resolve reinvite retries once

charliesantos:
**Contributing to Twilio**

### JIRA link(s):

- [CLIENT-6335](https://issues.corp.twilio.com/browse/CLIENT-6335)

### Description

Switching to events instead of promise to fix a bug where answer is applied multiple times after creating an offer

### Before review
* [x] Ready for review

### Before merge
* [x] Got one or more +1s
* [x] Re-tested if necessary


charliesantos:
@ryan-rowland , I'm thinking of changing this
https://github.com/twilio/twilio-client.js/blob/CLIENT-6335/lib/twilio/rtc/peerconnection.js#L699
to a warning instead since the we retry anyways. This might become an issue if customer is doing something with this. wdyt?

lib/twilio/rtc/peerconnection.js
```diff
@@ -680,7 +681,15 @@ PeerConnection.prototype.iceRestart = function() {
   return new Promise((resolve, reject) => {
     this.log('Attempting to restart ICE...');
     this.version.createOffer(this.codecPreferences, { iceRestart: true }).then(() => {
-      this.pstream.reinvite(this.version.getSDP(), this.callSid).then(payload => {
+      const removeListeners = () => {
```
ryan-rowland: Let's move this to the PeerConnection prototype so we don't have to keep defining it. Something like `this._removeReconnectionListeners`. We also want to make sure we call it when we clean up PeerConnection with `PeerConnection.prototype.close` too. Right now we just remove the _onAnswerOrRinging handler but not the new _onHangup answer.

charliesantos:
Received +1. Merging now to kick off travis on master for rc2

PR #20, opened 2019-07-03
CLIENT-6341 CLIENT-6342 | Ignore warnings if total bytes is zero. Disconnect on signaling failed

charliesantos:
**Contributing to Twilio**

### JIRA link(s):

- [CLIENT-6341](https://issues.corp.twilio.com/browse/CLIENT-6341)
- [CLIENT-6342](https://issues.corp.twilio.com/browse/CLIENT-6342)

### Description

- do not trigger ice restart if total bytes is zero
- when websocket gets disconnected, we also disconnect the current call since there's no way to resume the call once websocket is disconnected.

**Details why it's happening only on Firefox:**
Firefox throws an error when calling createOffer during iceRestart when you lose internet connection. We rely on the server to return a 'hangup' event to stop retrying. Since firefox won't get passed the createOffer step, we will never get a response back from the server.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/lib/twilio/connection.ts
+++ b/lib/twilio/connection.ts
@@ -272,7 +272,7 @@ class Connection extends EventEmitter {
 
         if (samples && samples.every(sample => sample.totals[name] === 0)) {
           // We don't have relevant samples yet, usually at the start of a call.
-          // On Edge browser, this is always zero
+          // This also may mean the browser does not support the required fields.
           return;
         }
 
@@ -297,7 +297,7 @@ class Connection extends EventEmitter {
 
         if (samples && samples.every(sample => sample.totals[name] === 0)) {
           // We don't have relevant samples yet, usually at the start of a call.
-          // On Edge browser, this is always zero
+          // This also may mean the browser does not support the required fields.
           return;
         }
 
--- a/lib/twilio/connection.ts
+++ b/lib/twilio/connection.ts
@@ -423,6 +423,10 @@ class Connection extends EventEmitter {
     this.pstream.on('cancel', this._onCancel);
     this.pstream.on('ringing', this._onRinging);
 
+    // When websocket gets disconnected
+    // There's no way to retry this session so we disconnect
+    this.pstream.on('offline', this._disconnect.bind(this));
+
     this.on('error', error => {
       this._publisher.error('connection', 'error', {
         code: error.code, message: error.message,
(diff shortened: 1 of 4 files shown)


lib/twilio/connection.ts
```diff
@@ -285,7 +292,15 @@ class Connection extends EventEmitter {
       this._reemitWarning(data, wasCleared);
     });
     monitor.on('warning-cleared', (data: RTCWarning) => {
-      if (data.name === 'bytesSent' || data.name === 'bytesReceived') {
+      const { samples, name } = data;
+      if (name === 'bytesSent' || name === 'bytesReceived') {
+
+        if (samples && samples.every(sample => sample.totals[name] === 0)) {
```
ryan-rowland: Is it possible to get here if all of the samples are zero?

lib/twilio/rtc/monitor.ts
```diff
@@ -380,18 +380,18 @@ class RTCMonitor extends EventEmitter {
     if (typeof limits.max === 'number') {
       count = countHigh(limits.max, values);
       if (count >= raiseCount) {
-        this._raiseWarning(statName, 'max', { values });
+        this._raiseWarning(statName, 'max', { values, samples: relevantSamples });
```
ryan-rowland: This is a breaking change. It breaks what is emitted via `Connection.on('warning')` and `Connection.on('warning-cleared')`

ryan-rowland: Same with other 3 line changes in this file.

lib/twilio/connection.ts
charliesantos: yes, we specify in rtc monitor a min threshold for current bytes, which is 1. If goes below one, 3 times, this gets emitted.

lib/twilio/rtc/monitor.ts
charliesantos: It's an additional data right? I thought Connection.on('warning') and Connection.on('warning-cleared') only cares about the `values` property?

lib/twilio/connection.ts
ryan-rowland: Ah I see. Right. We're triggering on current but checking total. 👍

charliesantos:
This is ready @ryan-rowland

ryan-rowland:
+1 to merge after the quick comment update

charliesantos:
Ryan is ok to merge. Merging..

PR #21, opened 2019-07-10
WIP - New Errors

ryan-rowland:
WIP @charliesantos 

This is the direction I'm headed, just wanted to check in with you on the approach before moving to adding tests and adding to code. Basically, this will be consumed in a second file like so:
```
import TwilioError from './error';

export const SignalingDisconnectedError: TwilioError = new TwilioError(53001);
export const ApplyLocalMediaFailedError: TwilioError = new TwilioError(53004);
```

And be thrown like so:
```
import { SignalingDisconnectedError, ApplyLocalMediaFailedError } from './knownErrors';

if (signalingDisconnected) {
  throw SignalingDisconnectedError.withMessage('A message giving more specific details');
}
```

Output will look something like:
```
{
  code: 53001,
  name: 'ConnectionDisconnected',
  description: 'Signaling connection disconnected',
  explanation: 'Raised whenever the signaling connection is unexpectedly disconnected.',
  message: 'A message giving more specific details',
  knownCauses: [{
    cause: 'The device running your application lost its Internet connection.',
    solution: 'Ensure the device running your application has access to a stable Internet connection.',
  }],
  withMessage: [Function]
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,9 @@ New Features
 This feature, when `enableIceRestart` is enabled, allows for detecting when media connection fails which will trigger automatic media reconnection, and for detecting when media connection is restored.
 
 #### New Events
-* `Connection.on('reconnecting', handler(error))` - raised when media connection fails and automatic reconnection has been started. Media connection failure triggers when there is no media flowing for approximately more than 3 seconds. During this period, `Connection.status()` will be set to `reconnecting`;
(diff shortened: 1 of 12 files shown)


lib/twilio/error.ts
```diff
@@ -0,0 +1,84 @@
...
+  causes: string[] | undefined;
+  code: number;
+  description: string;
+  explanation: string;
+  name: string;
+  solutions: string[] | undefined;
+
+  constructor(code: number) {
+    super();
+    Object.assign(this, builtErrors.get(code), { code });
```
charliesantos: Should we have some logic if `code` doesn't exist in builtErrors?

charliesantos:
I think it looks good. I just left one question.

charliesantos:
Also, don't forget to submit this PR against 2.0 branch instead :)

PR #23, opened 2019-07-12
CLIENT-6361 | Add better warning when WebRTC does not work due to http

ryan-rowland:


lib/twilio/device.ts
```diff
@@ -532,6 +532,12 @@ class Device extends EventEmitter {
    */
   setup(token: string, options: Device.Options = { }): this {
     if (!Device.isSupported && !options.ignoreBrowserSupport) {
+      if (window && window.location && window.location.protocol === 'http:') {
```
charliesantos: Should we put this check on device setup in general? that is, move it outside of `if (!Device.isSupported && !options.ignoreBrowserSupport)`?

ryan-rowland: Nope, we only want to throw an error if the browser isn’t letting us use WebRTC. Sometimes using this over http works, like on localhost.

PR #24, opened 2019-07-12
Client 6383 client 6372 | Listening to websocket close and checking for signaling state before applying answer

charliesantos:
**Contributing to Twilio**

### JIRA link(s):

- [CLIENT-6383](https://issues.corp.twilio.com/browse/CLIENT-6383)
- [CLIENT-6372](https://issues.corp.twilio.com/browse/CLIENT-6372)

### Description

- We were using `pstream.on('offline')` to detect network drops which would disconnect the active connection. But signaling server also emits 'offline' event when the token expires, which causes unwanted disconnection. This PR switches to `transport.on('close')` instead which gets emitted on websocket close.
- We are now ignoring iceRestart if signaling state is not `have-local-offer`, which usually happens when a createOffer came in late and a previous answer has already been applied.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/lib/twilio/connection.ts
+++ b/lib/twilio/connection.ts
@@ -425,7 +425,7 @@ class Connection extends EventEmitter {
 
     // When websocket gets disconnected
     // There's no way to retry this session so we disconnect
-    this.pstream.transport.on('close', this._disconnect.bind(this));
+    this.pstream.on('transportClosed', this._disconnect.bind(this));
 
     this.on('error', error => {
       this._publisher.error('connection', 'error', {
--- a/lib/twilio/connection.ts
+++ b/lib/twilio/connection.ts
@@ -425,7 +425,7 @@ class Connection extends EventEmitter {
 
     // When websocket gets disconnected
     // There's no way to retry this session so we disconnect
-    this.pstream.on('offline', this._disconnect.bind(this));
+    this.pstream.transport.on('close', this._disconnect.bind(this));
 
     this.on('error', error => {
       this._publisher.error('connection', 'error', {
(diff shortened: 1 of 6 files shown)


lib/twilio/connection.ts
```diff
@@ -425,7 +425,7 @@ class Connection extends EventEmitter {
 
     // When websocket gets disconnected
     // There's no way to retry this session so we disconnect
-    this.pstream.on('offline', this._disconnect.bind(this));
+    this.pstream.transport.on('close', this._disconnect.bind(this));
```
ryan-rowland: We shouldn’t be reaching into a child here. This means Connection will know about pstream’s implementation details. Maybe we should add a new event to pstream, `close`. WDYT?

lib/twilio/rtc/peerconnection.js
```diff
@@ -698,6 +698,13 @@ PeerConnection.prototype.iceRestart = function() {
         this._removeReconnectionListeners();
         if (!payload.sdp) { return reject(true); }
 
+        // Either this did not come from createOffer from iceRestart
+        // or a createOffer came in late and a previous anwer has already been applied
```
ryan-rowland: sp: answer

lib/twilio/connection.ts
charliesantos: Good point, however, pstream already has its own 'close' event which has a different meaning. It's the 'close' event coming from the server, which doesn't necessarily mean websocket's 'close'.

charliesantos: Or, we can still add a close event in pstream and call it something different. Something that's related to transport close, example `transport-close`. WDYT?

ryan-rowland: Man, we should rethink the way we surface these messages. Separate messages received from transport state. Let’s go with `transportClosed` for now.

charliesantos: I agree @ryan-rowland .

PR #35, opened 2019-07-24
CLIENT-6379 | Add error generation script

ryan-rowland:
@charliesantos This adds a build step that compiles a typescript file containing all errors we plan to use in the SDK. Still a little work to do but mostly complete. I just need to:
- Move the declaration of the errors we will use out of the script and into a settings file and
- Figure out why the 4 test assertions are failing. For some reason these are not recognized as `instanceof` their parent error types, except `Error` itself.


charliesantos:
@ryan-rowland this looks good to me. Should we also publish @twilio/voice-errors to npm?

charliesantos:
instanceof seems broken for classes that extends Error object in ts.
https://github.com/Microsoft/TypeScript/issues/13965

PR #38, opened 2019-07-29
Part 1 adding errors: device and connection

charliesantos:
Adding errors for device and connection. Errors not captured this this PR are listed and being discussed here https://paper.dropbox.com/doc/Voice-Errors-Updates--Ah2OY4BoQMQw6xCUloVwEpP8Ag-zOyPX4RE6C9gdSQddWlww

Here's a file matrix to see progress: https://docs.google.com/spreadsheets/d/17OW2ErW4HmdZj0eChkMjDHlW1nlKsHZYe8Pea1MAFW4/edit#gid=0

PR #40, opened 2019-07-29
Part 2 adding errors: pstream and wstransport

charliesantos:
Adding errors to pstream and wstransport

Here's a file matrix to see progress: https://docs.google.com/spreadsheets/d/17OW2ErW4HmdZj0eChkMjDHlW1nlKsHZYe8Pea1MAFW4/edit#gid=0


ryan-rowland:
https://github.com/twilio/twilio-client.js/blob/ae16a1e319dacccf75b41ba68f90f10fab7809ec/lib/twilio/pstream.js#L140

I think this can be `error`. We should attach a general signaling TwilioError here in the case that this event is `error`. I think we'll need to make a new error for this.

charliesantos:
Merge https://github.com/twilio/twilio-client.js/pull/38 first before merging this PR.

PR #46, opened 2019-08-07
CLIENT-6444 | Connection Management

charliesantos:
**Contributing to Twilio**

### JIRA link(s):

- [CLIENT-6444](https://issues.corp.twilio.com/browse/CLIENT-6444)

### Description

Implemented media connection management base on https://docs.google.com/document/d/1Ul1Z0yFOAUv8iopB6PQ0PsrWp8jrEz-T67YbBr9jt60/edit#heading=h.sh61l278o4f7

JS-related summary is here https://issues.corp.twilio.com/browse/CLIENT-6444

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/lib/twilio/connection.ts
+++ b/lib/twilio/connection.ts
@@ -12,8 +12,9 @@ import RTCMonitor from './rtc/monitor';
 import RTCSample from './rtc/sample';
 import RTCWarning from './rtc/warning';
 import Log, { LogLevel } from './tslog';
-import { average, Exception } from './util';
+import { average, isChrome } from './util';
 
+const Backoff = require('backoff');
 const C = require('./constants');
 const PeerConnection = require('./rtc').PeerConnection;
 
@@ -39,13 +40,28 @@ export type IPublisher = any;
  */
 export type ISound = any;
 
+const BACKOFF_CONFIG = {
+  factor: 1.1,
+  initialDelay: 1,
+  maxDelay: 30000,
+  randomisationFactor: 0.5,
+};
+
 const DTMF_INTER_TONE_GAP: number = 70;
 const DTMF_PAUSE_DURATION: number = 500;
 const DTMF_TONE_DURATION: number = 160;
 
-const ICE_RESTART_INTERVAL = 3000;
 const METRICS_BATCH_SIZE: number = 10;
 const METRICS_DELAY: number = 5000;
+
+const MEDIA_DISCONNECT_ERROR = {
+  disconnect: true,
+  info: {
+    code: 31003,
+    message: 'Connection with Twilio was interrupted.',
+    twilioError: new MediaErrors.ConnectionError(),
+  },
+};
 
 const WARNING_NAMES: Record<string, string> = {
   audioInputLevel: 'audio-input-level',
@@ -126,11 +142,6 @@ class Connection extends EventEmitter {
   private readonly _direction: Connection.CallDirection;
 
   /**
-   * Interval id for ICE restart loop
-   */
-  private _iceRestartIntervalId: NodeJS.Timer;
-
-  /**
    * The number of times input volume has been the same consecutively.
    */
   private _inputVolumeStreak: number = 0;
@@ -171,6 +182,16 @@ class Connection extends EventEmitter {
   private readonly _log: Log = new Log(LogLevel.Off);
 
   /**
+   * An instance of Backoff for media reconnection
+   */
+  private _mediaReconnectBackoff: any;
+
+  /**
+   * Timestamp for the initial media reconnection
+   */
+  private _mediaReconnectStartTime: number;
+
+  /**
    * A batch of metrics samples to send to Insights. Gets cleared after
    * each send and appended to on each new sample.
    */
@@ -200,6 +221,11 @@ class Connection extends EventEmitter {
    * State of the {@link Connection}.
    */
   private _status: Connection.State = Connection.State.Pending;
+
+  /**
+   * State of the {@link Connection} prior media reconnection.
+   */
+  private _statusPriorMediaFailure: Connection.State;
 
   /**
    * TwiML params for the call. May be set for either outgoing or incoming calls.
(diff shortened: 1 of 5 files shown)


lib/twilio/connection.ts
```diff
@@ -201,6 +222,11 @@ class Connection extends EventEmitter {
    */
   private _status: Connection.State = Connection.State.Pending;
 
+  /**
+   * State of the {@link Connection} prior media reconnection.
+   */
+  private _statusPriorMediaFailure: Connection.State;
```
ryan-rowland: `_statusPriorToMediaFailure`? `_statusBeforeFailure`?

Is there a way we can achieve the same thing without needing to store this? Maybe we keep `_status` the same (don't transition to reconnecting) and instead have an `_isReconnecting` flag? Then we can change the `status()` method to return `reconnecting` if it's true. Just a thought. WDYT?

lib/twilio/connection.ts
```diff
@@ -254,6 +280,9 @@ class Connection extends EventEmitter {
       : this.options.warnings ? LogLevel.Warn
       : LogLevel.Off);
 
+    this._mediaReconnectBackoff = Backoff.exponential({...BACKOFF_CONFIG});
```
ryan-rowland: Why are we copying the backoff config object? This method doesn't mutate it, does it?

lib/twilio/connection.ts
```diff
@@ -368,20 +362,27 @@ class Connection extends EventEmitter {
...
       this._log.info(msg);
       this._publisher.warn('network-quality-warning-raised', 'ice-connectivity-lost', {
         message: msg,
       }, this);
       this.emit('warning', 'ice-connectivity-lost');
+
+      this._onMediaFailure(Connection.MediaFailure.IceConnectionDisconnected);
+    };
+
+    this.mediaStream.onicefailure = (msg: string): void => {
```
ryan-rowland: Should this be `onicefailed`?

lib/twilio/connection.ts
```diff
@@ -1150,6 +1142,103 @@ class Connection extends EventEmitter {
...
+      if (type === IceConnectionFailed || type === PcConnectionFailed) {
+
+        // We already exceeded max retry time.
+        if (Date.now() - this._mediaReconnectStartTime > BACKOFF_CONFIG.maxDelay) {
+          this._log.info('Exceeded max ICE retries');
+          return this.mediaStream.onerror(MEDIA_DISCONNECT_ERROR);
+        }
+
+        // Triggers the next retry. This throws an error if multiple backoffs are called simultaneously
+        // This is the case for browsers which supports pc.connectionState and hasn't deprecated pc.iceConnectionState
```
ryan-rowland: Is there a better way to detect this? Are we able to just use one or the other (iceConnectionState or connectionState) instead of both? Can we detect which is supported up front and only attach to one?

lib/twilio/rtc/monitor.ts
```diff
@@ -204,6 +204,17 @@ class RTCMonitor extends EventEmitter {
     return this;
   }
 
+  /**
+   * Check if there is an active warning for a specific stat and threshold
+   * @param statName - The name of the stat to check
+   * @param thresholdName - The name of the threshold to check
+   * @returns Whether there is an active warning for a specific stat and threshold
+   */
+  hasWarning(statName: string, thresholdName: string): boolean {
```
ryan-rowland: How about `hasActiveWarning`? So it's a bit clearer.

lib/twilio/rtc/peerconnection.js
```diff
@@ -694,54 +694,51 @@ PeerConnection.prototype._removeReconnectionListeners = function() {
...
+    };
 
-    }, () => reject(true));
+    this._onHangup = () => {
+      this.log('Received hangup during ICE Restart');
+      this._removeReconnectionListeners();
+    };
+
+    this.pstream.on('answer', this._onAnswerOrRinging);
+    this.pstream.on('hangup', this._onHangup);
```
ryan-rowland: Is this handler being unset during cleanup?

lib/twilio/connection.ts
charliesantos: Good idea. I think that is possible. I will update.

lib/twilio/connection.ts
charliesantos: Correct it doesn't mutate it. It's out of habit and I don't think that applies here :)

lib/twilio/connection.ts
charliesantos: I'm following how we're naming other events in this file e.g. `onicereconnect`, `onicedisconnect` which are present tense. I can make them past tense.

lib/twilio/connection.ts
charliesantos: I tried 😄 . The problem is, some browsers doesn't fully implement either. For example, chrome supports both but  only `connectionState` transitions correctly but `iceConnectionState` triggers more reliably on `disconnect`.
Safari supports both but none transitions to `connected` after ice restart. Firefox and edge are not reliable, they switch to 'failed' faster (but not applicable in this comment).

If we want to use one or the other, I ended up a harder to read code around this area https://github.com/twilio/twilio-client.js/blob/master/lib/twilio/rtc/peerconnection.js#L616. I think the current implementation is the best compromise. But I can try to change it if you think it's worth it.

lib/twilio/rtc/peerconnection.js
charliesantos: Yes, they are being cleaned up inside `_removeReconnectionListeners()`, which is called on `PeerConnection.close`, and also before and after `PeerConnection.iceRestart` is complete.

lib/twilio/connection.ts
ryan-rowland: Ah. I thought they were being named based on the state. Maybe we should use the state name.

lib/twilio/connection.ts
charliesantos: Actually, this is not needed anymore because we always transition to `open`. I'll remove this.

PR #51, opened 2019-08-14
CLIENT-6336 | Fix infinite ringing on network change

mhuynh5757:
**Contributing to Twilio**

### JIRA link(s):

- [CLIENT-6336](https://issues.corp.twilio.com/browse/CLIENT-6336)

### Description

Fixes the following observation:
* A has a pending call from B.
* A undergoes a network change (i.e. disconnect and reconnect).
* The call is canceled for B, but exists indefinitely as `pending` for A.
* A hears the line ringing indefinitely.

This PR changes the behavior of `Device` when it witnesses a `transportClose` event emitted by a `Connection`. It will now remove and disconnect the `Connection` if it is pending while a `transportClose` event occurs. This stops the indefinite ringing.

### Before review
* [ ] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


lib/twilio/connection.ts
```diff
@@ -425,18 +425,15 @@ class Connection extends EventEmitter {
...
-        this._disconnect(info.message);
-
-        this._log.error('Received transportClose from pstream', info);
-        this.emit('error', {...info, connection: this});
-      });
-    }
+    // When websocket gets disconnected
+    // There's no way to retry this session so we disconnect
+    this.pstream.on('transportClose', () => {
+      const info = {...MEDIA_DISCONNECT_ERROR.info};
```
charliesantos: One thing we forgot to handle. Please add a check here so that we don't call disconnect again and we don't emit this error if the connection is already disconnected. Or else this error will get emitted multiple times because we are reconnecting wstransport. I understand this is happening on 1.8 branch, but only with ice restart enabled. This will become more apparent now that we're removing this `if` check.

charliesantos: Something like
```
this.pstream.on('transportClose', () => {
  if (I'm already disconnected) {
    return;
  }
  // Your rest of the stuff here
});
```

charliesantos:
These are the current behaviors before this fix. These issues happen all the way back to 1.7.1 and has been addressed if `enableIceRestart` flag is set to true. However, by default, this flag is set to false.

| Issue | Steps | Chrome | Firefox | Safari | Workaround |
| --- | --- | --- | --- | --- | --- |
|**1)** Cannot end call, cannot make future calls, but can still hear other end (No JIRA) | Switch network while a call is in progress | X |  | X | refresh browser |
|**2)** Cannot hear other end. Call will automatically disconnect after 15-20 seconds  (No JIRA) | Switch network while a call is in progress |  | X |   | waith 15-20 seconds |
|**3)** Infinite ringing sound (CLIENT-6366) | Disconnect network while ringing | X | X | X | refresh browser |

With @mhuynh5757 's fix, since we cannot recover the signaling connection, we will remove the `if(enableIceRestart)` check in [one area of the code](https://github.com/twilio/twilio-client.js/pull/51/files#diff-2ec8416c9cc5a81e6e7db4b6d8966f51L428), which will disconnect the call immediately when we detect a network handoff/handover, this would address all issues above. However, we have a concern on the first issue above. We are basically "failing" now by trying to disconnect, instead of just letting the call to keep going until the user realize that they need to refresh the browser. Is this ok? @ryan-rowland @aymenn

lib/twilio/connection.ts
```diff
@@ -425,18 +425,10 @@ class Connection extends EventEmitter {
...
-    if (this.options.enableIceRestart) {
-      // When websocket gets disconnected
-      // There's no way to retry this session so we disconnect
-      // This is not needed if ice restart is disabled, signaling will automatically disconnect the connection
-      this.pstream.on('transportClose', () => {
-        const info = {...MEDIA_DISCONNECT_ERROR.info};
-        this._disconnect(info.message);
-
-        this._log.error('Received transportClose from pstream', info);
-        this.emit('error', {...info, connection: this});
```
charliesantos: This is now handle inside _onMediaFailure.

PR #52, opened 2019-08-16
Switch maxBitrate to use maxaveragebitrate via SDP munging

ryan-rowland:
New description:
This implements the `maxaveragebitrate` approach of applying maxBitrate by munging the SDP. This negotiates a bitrate to use at both ends of the call, and is no longer limited to only the outgoing bitrate.

Original description:
> https://issues.corp.twilio.com/browse/CLIENT-5589
> 
> Apparently lower bound on Opus is 6kpbs and Chrome limits to over 12kpbs. So if a customer passes lower than 6kpbs they get an error and under 12kpbs it is silently ignored. Added a check for under 12001 and set minimum to 12001.
> 
> Also added a check for PCMU being set because we get an error if we try to set a bitrate for PCMU.

### DIFF
--- a/lib/twilio/rtc/sdp.js
+++ b/lib/twilio/rtc/sdp.js
@@ -5,6 +5,19 @@ const ptToFixedBitrateAudioCodecName = {
   8: 'PCMA'
 };
 
+const defaultOpusId = 111;
+
+function setMaxBitrate(sdp, maxBitrate) {
+  const matches = /a=rtpmap:(\d+) opus/gm.exec(sdp);
+  const opusId = matches && matches.length ? matches[1] : defaultOpusId;
+  const regex = new RegExp(`a=fmtp:${opusId}`);
+  const lines = sdp.split('\n').map(line => regex.test(line)
+    ? line + `;maxaveragebitrate=${maxBitrate}`
+    : line);
+
+  return lines.join('\n');
+}
+
 /**
  * Return a new SDP string with the re-ordered codec preferences.
  * @param {string} sdp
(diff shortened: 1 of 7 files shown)


lib/twilio/rtc/sdp.js
```diff
@@ -5,6 +5,14 @@ const ptToFixedBitrateAudioCodecName = {
   8: 'PCMA'
 };
 
+function setMaxBitrate(sdp, maxBitrate) {
+  const lines = sdp.split('\n').map(line => /a=fmtp:111/.test(line)
```
manjeshbhargav: @ryan-rowland Are you sure the the payload type for opus is always 111?

ryan-rowland: I'll add some additional logic similar to C++ to grab from the `opus` line just to be sure. 👍

charliesantos:
I'll let @manjeshbhargav check the SDP munging part. Other parts of the code looks good. Should we update changelog to include minimum (and possibly maximum) value for `maxBitrate`?

lib/twilio/rtc/sdp.js
```diff
@@ -5,6 +5,19 @@ const ptToFixedBitrateAudioCodecName = {
   8: 'PCMA'
 };
 
+const defaultOpusId = 111;
+
+function setMaxBitrate(sdp, maxBitrate) {
```
manjeshbhargav: @ryan-rowland You can simplify this to:
```js
function setMaxBitrate(sdp, maxBitrate) {
  const [opusPayloadType] = sdp.match(/a=rtpmap:(\d+) opus/) || [defaultOpusPayloadType];
  const opusFmtpRegex = new RegExp(`a=fmtp:${opusPayloadType} (.+)`, 'gm');
  return sdp.replace(opusFmtpRegex, `a=fmtp:${opusPayloadType} $1;maxaveragebitrate=${maxBitrate}`);
}
```
WDYT?

ryan-rowland:
Pulled master and opened against master here https://github.com/twilio/twilio-client.js/pull/55

PR #54, opened 2019-08-22
CLIENT-6539 | Getting audio levels and add insights for extensions

charliesantos:
**Contributing to Twilio**

### JIRA link(s):

- [CLIENT-6539](https://issues.corp.twilio.com/browse/CLIENT-6539)

### Description

- Get audio levels without relying on requestAnimationFrame which won't trigger when a tab is inactive or when running as an extension.
- Adding insights whether running as an extension.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+1.8.1 (In Progress)
+====================
+
+Bug Fixes
+---------
+
+* Fixed an issue where audio levels reported are zero when running as an extension, or when the browser tab is inactive or minimized.
+
+
 1.8.0 (Aug 20, 2019)
 ====================
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 Bug Fixes
 ---------
 
-* Fixed an issue where audio levels reported are zero when running as an extension, or when the browser tab is inactive or minimized.
+* Fixed an issue causing audio levels to be reported as zero when running as an extension, or when the browser tab is inactive or minimized.
 
 
 1.8.0 (Aug 20, 2019)
--- a/lib/twilio/connection.ts
+++ b/lib/twilio/connection.ts
@@ -1380,10 +1380,8 @@ namespace Connection {
   declare function rejectEvent(connection: Connection): void;
 
   /**
-   * Emitted on `requestAnimationFrame` (up to 60fps, depending on browser) with
-   *   the current input and output volumes, as a percentage of maximum
-   *   volume, between -100dB and -30dB. Represented by a floating point
-   *   number.
+   * Emitted up to 20hz with the current input and output volumes, as a percentage of maximum
+   * volume, between -100dB and -30dB. Represented by a floating point number.
    * @param inputVolume - A floating point number between 0.0 and 1.0 inclusive.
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+1.8.1 (In Progress)
+====================
+
+Bug Fixes
+---------
+
+* Fixed an issue where audio levels reported are zero when running as an extension, or when the browser tab is inactive or minimized.
```
ryan-rowland: "Fixed an issue causing audio levels to be reported as zero when running as an extension, or when the browser tab is inactive or minimized"

lib/twilio/connection.ts
```diff
@@ -1380,10 +1380,8 @@ namespace Connection {
   declare function rejectEvent(connection: Connection): void;
 
   /**
-   * Emitted on `requestAnimationFrame` (up to 60fps, depending on browser) with
-   *   the current input and output volumes, as a percentage of maximum
-   *   volume, between -100dB and -30dB. Represented by a floating point
-   *   number.
+   * Emitted up to 20hz with the current input and output volumes, as a percentage of maximum
```
ryan-rowland: I don't think `hz` is common outside of monitor refresh rates, is it? How about `Emitted every 50ms`?

lib/twilio/device.ts
```diff
@@ -277,6 +277,11 @@ class Device extends EventEmitter {
     [Device.SoundName.Outgoing]: true,
   };
 
+  /**
+   * Whether SDK is run as a browser extension
+   */
+  private _isBrowserExtension: boolean;
```
ryan-rowland: Does this catch extensions for all browsers or just Chrome?

lib/twilio/device.ts
```diff
@@ -797,6 +808,7 @@ class Device extends EventEmitter {
    */
   private _createDefaultPayload = (connection?: Connection): Record<string, any> => {
     const payload: Record<string, any> = {
+      browser_extension: this._isBrowserExtension,
```
ryan-rowland: Prefer `is_extension`. We will need to add this to the Insights schema as well.

lib/twilio/rtc/peerconnection.js
```diff
@@ -8,6 +8,8 @@ const Log = require('../log');
 const util = require('../util');
 const RTCPC = require('./rtcpc');
 
+// 20hz
```
ryan-rowland: Nit: Let's remove this comment.

lib/twilio/rtc/peerconnection.js
```diff
@@ -8,6 +8,8 @@ const Log = require('../log');
...
+const EMIT_VOLUME_TIMEOUT = 50;
```
ryan-rowland: Prefer `VOLUME_INTERVAL_MS`

lib/twilio/connection.ts
charliesantos: I was thinking in terms of frequency. But yes, `in ms` would be clearer.

lib/twilio/device.ts
charliesantos: It will catch chrome, edge, and firefox. Safari doesn't support the same extension model.

PR #55, opened 2019-08-23
Implement maxaveragebitrate via SDP munging

ryan-rowland:


lib/twilio/connection.ts
```diff
@@ -1593,6 +1593,13 @@ namespace Connection {
      */
     getSinkIds?: () => string[];
 
+    /**
+     * Max bitrate, in bits per second, for the local audio stream. Only works with Opus as
+     * this parameter is not supported by PCMU, which has a fixed bitrate. The minimum
+     * bitrate supported by Opus is 6kbit/s.
+     */
+    maxBitrate?: number;
```
idelgado: For consistency across mobile it may be worthwhile to call this "maxAverageBitrate" https://twilio.github.io/twilio-voice-android/docs/4.3.0/com/twilio/voice/OpusCodec.html. This corresponds to the term used in the Opus RFC and in the SDP.

idelgado: And you may want to consider using this doc: https://twilio.github.io/twilio-voice-android/docs/4.3.0/com/twilio/voice/OpusCodec.html#getMaxAverageBitrate--

idelgado: You may also want to apply the same logic for how its applied. 

"By default, the setting is not used. If you specify 0, then the setting is not used. Any positive integer is allowed, but values outside the range 6000 to 510000 are ignored and treated as 0."

ryan-rowland: @idelgado I can see doing that for the sake of consistency, but I'm curious why we went with treating outside the range as zero rather than the upper or lower bound.

idelgado: Another way of saying this is "Values outside the supported range 6000 to 510000 are ignored and no max average bitrate is specified". I think written this way the above makes more sense.

PR #62, opened 2019-09-05
Add the settings:codec event after remote answer is applied

ryan-rowland:
This adds the following event based off of the [mobile implementation](https://github.com/twilio/rtc-cpp/pull/1105):
```js
group: settings
name: codec
payload:
  codec_params: string // eg: 'maxaveragebitrate=12000;usedtx=1'
  selected_codec: string // eg: 'opus/48000/2'
```

If either or neither of the codec name or codec params are found, an empty string will be reported in its place.

This uses the SDP we get after applying the remote answer, for both incoming and outgoing calls.

### DIFF
--- a/lib/twilio/connection.ts
+++ b/lib/twilio/connection.ts
@@ -16,7 +16,8 @@ import { average, isChrome } from './util';
 
 const Backoff = require('backoff');
 const C = require('./constants');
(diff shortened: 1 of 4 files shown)


charliesantos:
I verified for OPUS and PCMU. Works as expected.

dipankadas:
Looks good to me.

PR #68, opened 2019-09-11
CLIENT-6632 | Don't always download ringtones

charliesantos:
**Contributing to Twilio**

### JIRA link(s):

- [CLIENT-6632](https://issues.corp.twilio.com/browse/CLIENT-6632)

### Description

Currently, ringtone media files are always downloaded whenever we play them. Audio elements are always being created as well. This PR caches the Audio elements to prevent re-downloading of media files. Cache is invalidated if we change sinkIds.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/lib/twilio/sound.js
+++ b/lib/twilio/sound.js
@@ -37,7 +37,7 @@ function Sound(name, url, options) {
 
   Object.defineProperties(this, {
     _activeEls: {
-      value: new Set()
+      value: new Map()
     },
     _Audio: {
       value: options.AudioPlayer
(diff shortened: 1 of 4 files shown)


lib/twilio/sound.js
```diff
@@ -126,8 +152,9 @@ Sound.prototype.stop = function stop() {
 
 /**
  * Start playing the sound. Will stop the currently playing sound first.
+ * If it exists, the audio element that was initialized for the sinkId will be used
  */
-Sound.prototype.play = function play() {
+Sound.prototype.play = function play(isMuted, shouldLoop) {
```
ryan-rowland: If we're only using these arguments internally to this class we should rename this to `_play`, so the public `play` method doesn't have these arguments on it. It would then call `_play` with no arguments.

PR #72, opened 2019-09-17
CLIENT-6541 | Move integration tests inside Docker

charliesantos:
**Contributing to Twilio**

### JIRA link(s):

- [CLIENT-6541](https://issues.corp.twilio.com/browse/CLIENT-6541)

### Description

Moving existing integration tests inside docker. **_Network tests is a different ticket._**

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


ryan-rowland:
@makarandp0 What do you think about moving all integration tests into Docker like this? Didn't you find some issues in Video when attempting to do the same? Just want to make sure we don't repeat any mistakes here.

charliesantos:
@ryan-rowland , I talked to @makarandp0 about the issues

> there were two issues when running in travis.
> 1. travis sometimes timed out  while building the image. I did not figured out what went wrong there.
> 2. some of our network tests failed. And this part I believe is problem in our SDK code and not test or travis.
Issue 1 is not a problem anymore for video and for us. I'm not rebuilding the image for every build. I don't see issue 2 on ours 😄

karma.conf.ts
```diff
@@ -39,6 +39,9 @@ module.exports = function(config: any) {
       ChromeWebRTC: {
         base: 'Chrome',
         flags: [
+          '--headless',
```
makarandp0: does this allow you to run your tests outside docker? Just want to make sure that we keep that configuration working for developers.

makarandp0: Suggestion: always running --headless may make it bit harder to enable debug mode - perhaps make it --headless based on a environment flag or something? we use separate  configuration one for docker and one for outside: https://github.com/twilio/twilio-video.js/blob/master/karma/makeconf.js#L56

tests/docker/scripts/run.sh
```diff
@@ -0,0 +1,28 @@
...
+BVER=$BVER
+ACCOUNT_SID=$ACCOUNT_SID
+AUTH_TOKEN=$AUTH_TOKEN
+APPLICATION_SID=$APPLICATION_SID
+API_KEY_SECRET=$API_KEY_SECRET
+API_KEY_SID=$API_KEY_SID
+END
+
+echo "Preparing container"
+docker create --env-file env.list --name twilio-client-network-test csantos28/twilio-client:1.0.0
```
makarandp0: whats csantos28 in there :) ?

tests/docker/Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+RUN apt-get update
+RUN apt-get install -y libasound2 libpango1.0-0 libxt6 wget bzip2 sudo
+RUN apt-get install -y libdbus-glib-1-2
+RUN apt-get install -y libgtk-3-0
+
+WORKDIR /app
+
+COPY . /app
+
+CMD ["./entry.sh"]
```
makarandp0: instead of having entry.sh baked into it - I find it  useful to have it created with a 
generic 
```
CMD ["bash"]
```
and override it from the the command line where you run the image.

tests/docker/scripts/run.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+# Start of containerized test, originating from host machine
+set -e
+echo "Preparing environment variables"
+cat > env.list <<END
+BVER=$BVER
```
makarandp0: looks like lot of these settings can go into Docker-compose file, eventually you would need one as the arguments and options grow. It will be handy when you want to map docker socket into the docker image. 

But this can come as part of your network test PR.

karma.conf.ts
charliesantos: Yes, they can still run outside docker. `npm run test:integration` should run outside docker environment.

charliesantos: Good suggestion. I can make the headless flagg configurable.

tests/docker/scripts/run.sh
charliesantos: I'm using my own docker hub account to pull down the image. Do we have one for our team?

tests/docker/Dockerfile
charliesantos: Sorry I'm not sure I understand. Do you have an example for videojs?

tests/docker/scripts/run.sh
charliesantos: Good suggestion :)

tests/docker/Dockerfile
charliesantos: Actually, that might be confusing (at least for someone who is not a shell expert like me 😄 ). What would be the advantage of doing this? I saw other suggestions online using something like `ENTRYPOINT ["./entry.sh"]`. wdyt?

tests/docker/scripts/run.sh
makarandp0: Ah, so you have built this minimal docker image and uploaded using your account - And now you use that image from travis-job. I guess this works, but ideally we should either 1) let travis build the image or 2) use some job that builds the image ( ideally including browsers )

It would be good to check with @syerrapragada to see what docker account we can use, and if we can use his job to build our images. But again this need not hold this PR. Just include some script to build the docker image as part of the PR.

tests/docker/Dockerfile
makarandp0: The advantage of not specifying entry.sh as entrypoint in image  is it makes it easy to experiment against the image when using locally. you can simply run the image interactively and run commands inside it. 

Whereas Travis (or whatever CI) can specify complicated script that it wants to run in side the image when starting it. 

I guess it makes it easy for us, and difficult for travis. (instead of other way around)

tests/docker/scripts/run.sh
charliesantos: ok sounds good.

tests/docker/Dockerfile
charliesantos: This is a great suggestion after looking at it in action :)
I updated the PR to support this.

charliesantos:
Got 👍 to merge.

PR #74, opened 2019-09-18
CLIENT-6664 | Fix rtt and mos calculation for browsers not supporting…

charliesantos:
… dtlsState

**Contributing to Twilio**

### JIRA link(s):

- [CLIENT-6664](https://issues.corp.twilio.com/browse/CLIENT-6664)

### Description

Some browsers such as Safari doesn't support dtls state in rtc stats. If this doesn't exists, we currently skip rtt calculation which cuases mos to return null. This PR fixes that.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/lib/twilio/rtc/stats.js
+++ b/lib/twilio/rtc/stats.js
@@ -108,9 +108,7 @@ function createRTCSample(statsReport) {
 
         break;
       case 'transport':
-        if (stats.dtlsState === 'connected') {
-          activeTransportId = stats.id;
-        }
+        activeTransportId = stats.id;
         break;
     }
   });
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+1.9.2 (In Progress)
+===================
+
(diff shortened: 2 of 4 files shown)


lib/twilio/rtc/stats.js
```diff
@@ -108,9 +108,7 @@ function createRTCSample(statsReport) {
 
         break;
       case 'transport':
-        if (stats.dtlsState === 'connected') {
-          activeTransportId = stats.id;
-        }
+        activeTransportId = stats.id;
```
ryan-rowland: Any potential side effects here?

charliesantos: I don't think so. Down below, we are properly checking the existense of active transport so it shouldn't cause an issue. I tested on chrome firefox and edge

charliesantos: tested on safari as well 😄

PR #76, opened 2019-09-19
CLIENT-6542 | Dockerized network tests

charliesantos:
**Contributing to Twilio**

### JIRA link(s):

- [CLIENT-6542](https://issues.corp.twilio.com/browse/CLIENT-6542)

### Description

Network failure, media failure, and reconnection tests.
Also copying dockerProxy for now until we decide how to move forward with opensourcing.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


tests/integration/reconnection.ts
```diff
@@ -0,0 +1,294 @@
...
+          done();
+        });
+      });
+      runDockerCommand('resetNetwork');
+    });
+  });
+});
+
+function expectEvent(eventName: string, emitter: EventEmitter) {
+  return new Promise(resolve => {
```
makarandp0: NIT: could be shorter with  
``` return new Promise(resolve => emitter.once(eventName, resolve)); ```

File: tests/integration/reconnection.ts [hunk 2]
```diff
@@ -0,0 +1,294 @@
...
+      runDockerCommand('resetNetwork');
+    });
+    after(() => {
+      destroyDevices();
+      runDockerCommand('resetNetwork');
+    });
+
+    it('should trigger device.offline', (done: CB) => {
+      const id = setTimeout(() => { throw new Error('offline not detected'); }, EVENT_TIMEOUT);
+      bindTestPerDevice((device: Device) => {
```
makarandp0: shouldn't `bindTestPerDevice ` this go before `it` ? Like:
`bindTestPerDevice((device: Device) => { it(...) }`

tests/integration/reconnection.ts
```diff
@@ -0,0 +1,294 @@
...
+    });
+
+    it('should disconnect call with error 31003', (done: CB) => {
+      const id = setTimeout(() => { throw new Error('Disconnect timeout exceeded') }, EVENT_TIMEOUT);
+      const doneCb = (conn: Connection) => {
+        clearInterval(id);
+        assert(conn.status() === Connection.State.Closed);
+        done();
+      };
+      bindTestPerConnection((conn: Connection) => {
```
makarandp0: shouldn't `bindTestPerConnection` take test as a function param. Otherwise it will end-up calling `done`  multiple times for a single test.

tests/integration/reconnection.ts
```diff
@@ -0,0 +1,294 @@
...
+            withError = true;
+            disconnected && doneCb(conn);
+          }
+        });
+      });
+    });
+
+    it('should trigger device.ready after network resumes', (done: CB) => {
+      const id = setTimeout(() => { throw new Error('ready not detected'); }, EVENT_TIMEOUT);
+      bindTestPerDevice((device: Device) => {
```
makarandp0: same as last comment about `bindTestPerDevice`

tests/integration/reconnection.ts
```diff
@@ -0,0 +1,294 @@
...
+
+  const setupDevices = (done: CB) => {
+    identity1 = 'id1-' + Date.now();
+    identity2 = 'id2-' + Date.now();
+    token1 = generateAccessToken(identity1);
+    token2 = generateAccessToken(identity2);
+    device1 = new Device();
+    device2 = new Device();
+
+    Promise.all([
```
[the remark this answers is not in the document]
charliesantos: sounds good 👍

tests/integration/reconnection.ts
charliesantos: Good question. We are not running tests on both devices and expecting both to succeed. We are binding the test function for both devices, then whoever gets the event first will finish and pass the test. The other device will then do nothing. The reason why it was setup this way is because we don't know which device will get disconnected first... I know it is not straight forward, but I'm open for better suggestions.

tests/integration/reconnection.ts
charliesantos: See my reply for `device` above. Same situation.

tests/integration/reconnection.ts
charliesantos: See my reply for device above. Same situation.

tests/integration/reconnection.ts
charliesantos: I added comment in the code that better explains this.

tests/integration/reconnection.ts
```diff
@@ -0,0 +1,307 @@
...
+      before(() => {
+        runDockerCommand('unblockMediaPorts');
+        return setupDevices();
+      });
+      after(() => {
+        destroyDevices();
+        runDockerCommand('unblockMediaPorts');
+      });
+
+      it('should not trigger reconnecting', () => {
```
makarandp0: instead of ignoring the promise returned by `blockMediaPorts` you could
``` 
      it('should not trigger reconnecting', async () => {
        const mediaPortBlocked = runDockerCommand('blockMediaPorts');
        const testPromise =  Promise.race([
          wait(10000),
          bindTestPerConnection((conn: Connection) => new Promise((resolve, reject) => {
            conn.on('reconnecting', () => reject(new Error('Detected reconnecting')));
          }));
          await mediaPortBlocked;
          await testPromise; 
        ]);
      });
```

tests/integration/reconnection.ts
```diff
@@ -0,0 +1,307 @@
...
+
+    before(() => {
+      options.enableIceRestart = true;
+    });
+
+    describe('and ICE connection fails', function() {
+      this.timeout(USE_CASE_TIMEOUT);
+
+      before(() => {
+        runDockerCommand('unblockMediaPorts');
```
makarandp0: same here. shouldn't ignore promise from `runDockerCommand`.

tests/integration/reconnection.ts
```diff
@@ -0,0 +1,307 @@
...
+        return setupDevices();
+      });
+      after(() => {
+        destroyDevices();
+        runDockerCommand('unblockMediaPorts');
```
makarandp0: you might want to await for `runDockerCommand` here too...otherwise it may intefere with other tests.

tests/integration/reconnection.ts
```diff
@@ -0,0 +1,307 @@
...
+    before(() => setupDevices());
+    after(() => {
+      destroyDevices();
+      runDockerCommand('resetNetwork');
+    });
+
+    it('should trigger device.offline', () => {
+      runDockerCommand('disconnectFromAllNetworks');
+      return Promise.race([
+        wait(EVENT_TIMEOUT).then(() => Promise.reject('offline not detected')),
```
makarandp0: you can convert this pattern to a function that takes a promise and timeout, and then rejects if timeout happens before the promise. 

something like: 
https://github.com/twilio/twilio-video.js/blob/master/test/lib/util.js#L451

tests/integration/reconnection.ts
```diff
@@ -0,0 +1,307 @@
...
+      });
+      after(() => {
+        destroyDevices();
+        runDockerCommand('unblockMediaPorts');
+      });
+
+      it('should trigger reconnecting', () => {
+        runDockerCommand('blockMediaPorts');
+        return Promise.race([
+          wait(EVENT_TIMEOUT).then(() => Promise.reject('reconnecting not detected')),
```
makarandp0: you can convert this pattern into a util function that takes a promise and timeout, and then rejects if timeout happens before the promise. 

something like: 
https://github.com/twilio/twilio-video.js/blob/master/test/lib/util.js#L451

that way the usage becomes easier to read like
```
    it('should trigger reconnecting', async () => {
      await runDockerCommand('blockMediaPorts');
      const testPromise = bindTestPerConnection((conn: Connection) => new Promise((resolve) => {
          conn.on('reconnecting', () => {
            assert(conn.status() === Connection.State.Reconnecting);
            resolve();
          });
        }))
      await waitFor(testPromise, EVENT_TIMEOUT);
    });    

```

tests/integration/reconnection.ts
```diff
@@ -0,0 +1,307 @@
...
+          });
+        }))
+      ]);
+    });
+
+    it('should trigger device.ready after network resumes', () => {
+      runDockerCommand('resetNetwork');
+      return Promise.race([
+        wait(EVENT_TIMEOUT).then(() => Promise.reject('ready not detected')),
+        bindTestPerDevice((device: Device) => new Promise((resolve) => {
```
makarandp0: I see that you are using this 
```
new Promise((resolve) => {
  device.on('ready', () => resolve());
}))
```
patten everywhere. Can you convert all those to use `expectEvent` that you wrote:

```
    it('should trigger device.ready after network resumes', () => {
      runDockerCommand('resetNetwork');
      return Promise.race([
        wait(EVENT_TIMEOUT).then(() => Promise.reject('ready not detected')),
        bindTestPerDevice((device: Device) => new Promise((resolve) => {
          device.on('ready', () => resolve());
        }))
      ]);
    });
 ```
becomes 

```
    it('should trigger device.ready after network resumes', () => {
      runDockerCommand('resetNetwork');
      return Promise.race([
        wait(EVENT_TIMEOUT).then(() => Promise.reject('ready not detected')),
        bindTestPerDevice((device: Device) => expectEvent('ready', device));
      ]);
    });
```

and better still:

```
    it('should trigger device.ready after network resumes', async () => {
      await runDockerCommand('resetNetwork');
      await waitFor(bindTestPerDevice((device: Device) => expectEvent('ready', device)), EVENT_TIMEOUT);
    });
```

File: tests/integration/reconnection.ts (hunk 6 above)
Comment (charliesantos) [reply tests/integration/reconnection.ts]: Not sure I understand. There's some syntax error. Are you suggesting mediaPortBlocked and testPromise to be inside the Promise.race?

File: tests/integration/reconnection.ts [hunk 12]
```diff
@@ -165,84 +157,44 @@ describe('Reconnection', function() {
...
-        return Promise.race([
-          wait(EVENT_TIMEOUT).then(() => Promise.reject('reconnecting not detected')),
-          bindTestPerConnection((conn: Connection) => new Promise((resolve) => {
-            conn.on('reconnecting', () => {
-              assert(conn.status() === Connection.State.Reconnecting);
-              resolve();
-            });
-          }))
-        ]);
+      it('should trigger reconnecting', async () => {
```
makarandp0: with `async` tests, you do not have to return promise if you `await`   
```
      it('should trigger reconnecting', async () => {
        await runDockerCommand('blockMediaPorts');
        await waitFor(bindTestPerConnection((conn: Connection) => expectEvent('reconnecting', conn)
          .then(() => assert(conn.status() === Connection.State.Reconnecting))), EVENT_TIMEOUT);
      });
```

makarandp0: same goes for all other `async` test functions that return promises.

charliesantos: Thanks @makarandp0 . I updated.

PR #77, opened 2019-09-26
Dont run unnecessary tests inside release tool

charliesantos:
We are currently running tests twice during build releases which causes travis to fail. However, the job still successfully creates the tag.

But now that we have docker tests, these failures when run via release tools will stop the whole script.

This PR made it so that we only run test once.


charliesantos:
@ryan-rowland is ok to merge.

PR #80, opened 2019-10-02
Copying workingdir instead of mounting to fix permission issues

charliesantos:
Recently, we had an issue where artifacts cannot be deleted which breaks release process. This happens because docker is taking over the files that are mounted. This PR will not mount the files, but copy instead.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,9 +1,0 @@
-.dockerignore
-Dockerfile
-node_modules
-package-lock.json
-dist
-es5
-docs
-tests/docker/proxy/node_modules
-tests/docker/proxy/package-lock.json
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,4 +43,6 @@ RUN echo "Installing Firefox: $BVER" \
 
 ENV FIREFOX_BIN=/application/firefox/firefox
 
+COPY . /app
+
 CMD ["bash"]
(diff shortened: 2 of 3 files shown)


makarandp0:
@charliesantos, The way `Video` solves this problem is our release tool does not use docker at all. docker integration tests and release are separate jobs for us.

but your solution looks good too.

ryan-rowland:
Curious if there are any known side effects here. Are we skipping a build step that might cause us to test cached results, miss changes and have a false negative for errors?

charliesantos:
Currently, the host machine shares build artifacts with docker container. I dont see issues since both uses the same OS. But you bring up a good point. It will be safe if docker and host uses build artifacts to prevent issues like this PR is trying to solve. I can make it so that docker rebuilds the artifacts. Wdyt? @ryan-rownland @makarandp0

ryan-rowland:
As long as the artifacts are being built on the machine I think it's OK if we only build it once and then copy it into docker. Only potential problem I see there is a mismatch between OS / environment in the docker container vs the host system that causes a dependency error.

charliesantos:
Ok cool. In that case, this PR is good. We dont build on different OS/versions so there wont be issues. I double checked the build steps on regular builds and release builds, we are not skipping any steps.

PR #81, opened 2019-10-02
CLIENT-6474 | Clean up event listeners in Device

ryan-rowland:
Part 1 of (probably 3) for [CLIENT-6474](https://issues.corp.twilio.com/browse/CLIENT-6474). Making sure listeners get cleaned up in Device.

### DIFF
--- a/lib/twilio/device.ts
+++ b/lib/twilio/device.ts
@@ -460,8 +460,20 @@ class Device extends EventEmitter {
     }
 
     if (this.stream) {
+      this.stream.removeListener('close', this._onSignalingClose);
+      this.stream.removeListener('connected', this._onSignalingConnected);
+      this.stream.removeListener('error', this._onSignalingError);
+      this.stream.removeListener('invite', this._onSignalingInvite);
+      this.stream.removeListener('offline', this._onSignalingOffline);
+      this.stream.removeListener('ready', this._onSignalingReady);
+
       this.stream.destroy();
       this.stream = null;
+    }
+
+    const connection = this.activeConnection();
+    if (connection) {
+      connection.removeListener('error', this._onConnectionError);
     }
 
     if (networkInformation) {
@@ -914,7 +926,8 @@ class Device extends EventEmitter {
 
     const connection = new this.options.connectionFactory(config, options);
 
-    connection.once('accept', () => {
+    const onceConnectionAccept = (): void => {
+      stopListening();
       this._removeConnection(connection);
       this._activeConnection = connection;
       if (this.audio) {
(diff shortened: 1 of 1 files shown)


lib/twilio/device.ts
```diff
@@ -460,10 +460,22 @@ class Device extends EventEmitter {
...
+      this.stream.removeListener('offline', this._onSignalingOffline);
+      this.stream.removeListener('ready', this._onSignalingReady);
+
       this.stream.destroy();
       this.stream = null;
     }
 
+    const connection = this.activeConnection();
+    if (connection) {
+      connection.removeListener('error', this._onConnectionError);
```
charliesantos: Should we do this for all `this.connections` and not just the activeConnection?

lib/twilio/device.ts
```diff
@@ -926,56 +939,64 @@ class Device extends EventEmitter {
...
+      connection.removeListener('reject', onceConnectionReject);
+      connection.removeListener('transportClose', onceTransportClose);
+    }
+
+    connection.addListener('error', this._onConnectionError);
+    connection.once('accept', onceConnectionAccept);
+    connection.once('cancel', onceConnectionCancel);
+    connection.once('disconnect', onceConnectionDisconnect);
+    connection.once('reject', onceConnectionReject);
+    connection.once('transportClose', onceTransportClose);
```
charliesantos: What do you think about just removing all listeners for this connection on disconnect? We don't reuse the connection anyway right? Something like
```
connection.on('disconnect', () => {
    connection.eventNames().forEach(name => connection.removeAllListeners(name));
});
```

lib/twilio/device.ts
```diff
@@ -989,6 +1010,21 @@ class Device extends EventEmitter {
     }
   }
 
+  /**
+   * Called when the active Connection emits an error event.
+   */
+  private _onConnectionError = (error: Connection.Error): void => {
```
charliesantos: It seems like we are using this handler for all connections we create. I can see issues if we have multiple connections and one activeConnection. Errors originating from other connections will always affect the activeConnection - it will get removed, and will emit an error unexpectedly.

lib/twilio/device.ts
ryan-rowland: Good catch

lib/twilio/device.ts
ryan-rowland: Ugh yeah. This is a pain in the butt, managing all these functions as they are bound to a specific `Connection` instance. It seems like they should be associated with the `Connection` itself instead of the Device, maybe.

charliesantos:
@ryan-rowland I'm closing this for now until we bring this back into the sprint.

PR #89, opened 2019-10-24
[CLIENT-6831] Remove region checking logic

mhuynh5757:
**Contributing to Twilio**

### JIRA link(s):

- [CLIENT-6831](https://issues.corp.twilio.com/browse/CLIENT-6831)

### Description

This PR removes region checking logic to allow for the use of free-strings as region shortcodes. At the moment, there is no difference in the signaling stack logic between a network loss error and an incorrect region error, meaning that if the websocket can not connect in either case, we have the same error (and consequentially the same reconnect logic).

### Before review
* [ ] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+1.9.5 (In Progress)
+===================
+
+Improvements
+---------
+
+* In order to support new regions without requiring SDK updates in the future, the SDK no longer locally verifies the region option passed to `Device.setup` options. As a result, if an invalid region is passed, a connection error will be emitted after the connection is attempted. We recommend always testing to make sure your specified region (if any) works before pushing any change to production. For a list of supported regions refer to "https://www.twilio.com/docs/voice/client/regions". (CLIENT-6831)
+
+
 1.9.4 (Oct 28, 2019)
 ===================
 
@@ -299,8 +308,8 @@ Browser Support
 * We are now using RTCP values for RTT where available. Initially, this will not affect Chrome
   because Chrome has [not yet implemented support](https://bugs.chromium.org/p/webrtc/issues/detail?id=9545).
   Additionally, having access to RTT will allow FireFox to calculate and report MOS, however FireFox is
-  currently affected by [a regression](https://bugzilla.mozilla.org/show_bug.cgi?id=1525341) causing 
-  jitter to be reported as 0, which will make MOS scores appear slightly better than they actually 
+  currently affected by [a regression](https://bugzilla.mozilla.org/show_bug.cgi?id=1525341) causing
+  jitter to be reported as 0, which will make MOS scores appear slightly better than they actually
   are until it's fixed.
 
 1.6.7 (Feb 12, 2019)
--- a/lib/twilio/device.ts
+++ b/lib/twilio/device.ts
@@ -17,7 +17,11 @@ import {
   SignalingErrors,
 } from './errors';
 import { PStream } from './pstream';
-import { getRegionShortcode, getRegionURI, Region } from './regions';
+import {
+  defaultRegion,
+  getRegionShortcode,
+  getRegionURI,
+} from './regions';
 import Log, { LogLevel } from './tslog';
 import { Exception, queryToJson } from './util';
 
@@ -300,7 +304,7 @@ class Device extends EventEmitter {
   /**
    * The region the {@link Device} is connected to.
    */
-  private _region: Region | string | null = null;
+  private _region: string | null = null;
 
   /**
    * The current status of the {@link Device}.
@@ -328,7 +332,7 @@ class Device extends EventEmitter {
     iceServers: [],
     noRegister: false,
     pStreamFactory: PStream,
-    region: Region.Gll,
+    region: defaultRegion,
     rtcConstraints: { },
     soundFactory: Sound,
     sounds: { },
@@ -643,9 +647,9 @@ class Device extends EventEmitter {
       this.sounds[eventName] = getOrSetSound.bind(null, eventName);
     });
 
-    const regionURI = getRegionURI(this.options.region, (newRegion: string) => {
-      this._log.warn(`Region ${this.options.region} is deprecated, please use ${newRegion}.`);
-    });
+    const regionURI = getRegionURI(this.options.region, newRegion => this._log.warn(
+      `Region ${this.options.region} is deprecated, please use ${newRegion}.`,
+    ));
 
     this.options.chunderw = `wss://${this.options.chunderw || regionURI}/signal`;
 
@@ -1473,7 +1477,7 @@ namespace Device {
     /**
      * The region code of the region to connect to.
      */
-    region?: Region;
+    region?: string;
 
     /**
      * An RTCConfiguration to pass to the RTCPeerConnection constructor.
--- a/lib/twilio/regions.ts
+++ b/lib/twilio/regions.ts
@@ -6,7 +6,8 @@ import { InvalidArgumentError } from './errors';
 import { Exception } from './util';
 
 /**
- * Valid deprecated regions
+ * Valid deprecated regions.
+ *
  * @private
  */
 export enum DeprecatedRegion {
@@ -20,7 +21,12 @@ export enum DeprecatedRegion {
 }
 
 /**
- * Valid current regions
+ * Valid current regions.
+ *
+ * @deprecated CLIENT-6831
+ * This is no longer used or updated for checking validity of regions in the
+ * SDK. We now allow any string to be passed for region. Invalid regions won't
+ * be able to connect, and won't throw an exception.
  */
 export enum Region {
   Au1 = 'au1',
@@ -50,7 +56,7 @@ export type ValidRegion = Region | DeprecatedRegion;
 /**
  * Deprecated regions. Maps the deprecated region to its equivalent up-to-date region.
  */
-const deprecatedRegions: Record<DeprecatedRegion, Region> = {
+export const deprecatedRegions: Record<DeprecatedRegion, Region> = {
   [DeprecatedRegion.Au]: Region.Au1,
   [DeprecatedRegion.Br]: Region.Br1,
   [DeprecatedRegion.Ie]: Region.Ie1,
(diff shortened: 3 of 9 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+1.9.5 (In Progress)
+===================
+
+Improvements
+---------
+
+* We now accept any string for the region option when performing `Device.setup`. (CLIENT-6831)
```
ryan-rowland: Seems a bit ambiguous. How about:

```
* In order to support new regions without requiring SDK updates in the future, the SDK no longer locally verifies the region option passed to `Device.setup` options. As a result, if an invalid region is passed, an exception won't be thrown and the Device will be unable to connect. We recommend always testing to make sure your specified region (if any) works before pushing any change to production.

@aymenn What do you think?
```

ryan-rowland: @aymenn What do you think?

lib/twilio/device.ts
```diff
@@ -643,9 +648,19 @@ class Device extends EventEmitter {
       this.sounds[eventName] = getOrSetSound.bind(null, eventName);
     });
 
-    const regionURI = getRegionURI(this.options.region, (newRegion: string) => {
-      this._log.warn(`Region ${this.options.region} is deprecated, please use ${newRegion}.`);
-    });
+    const {
+      deprecated: regionDeprecated,
```
ryan-rowland: isRegionDeprecated

lib/twilio/device.ts
```diff
@@ -643,9 +648,19 @@ class Device extends EventEmitter {
...
+      newRegion,
+    } = checkRegionDeprecation(this.options.region);
```
ryan-rowland: Actually, can we merge this function (checkRegionDeprecation) and regionToChunderURI back into `getRegionURI`? It can return the value of regionToChunderURI if not deprecated region, and return regionToChunderURI(checkRegionDeprecation().newRegion if so.

lib/twilio/device.ts
```diff
@@ -643,9 +648,19 @@ class Device extends EventEmitter {
...
+
+    if (regionDeprecated) {
```
ryan-rowland: We can also expose another function from Region, `isRegionDeprecated(region) => boolean` for this part.

lib/twilio/regions.ts
```diff
@@ -7,6 +7,10 @@ import { Exception } from './util';
 
 /**
  * Valid deprecated regions
+ *
+ * NOTE(mhuynh | CLIENT-6831): This is no longer currently used for checking
+ * validity of regions in the SDK.
+ *
```
ryan-rowland: Is this true? I'm seeing us use it above to detect whether a region is deprecated, and throw a warning. I'd remove this comment addition.

lib/twilio/regions.ts
```diff
@@ -20,7 +24,10 @@ export enum DeprecatedRegion {
 }
 
 /**
- * Valid current regions
+ * Valid current regions.
+ *
+ * NOTE(mhuynh | CLIENT-6831): This is no longer currently used for checking
+ * validity of regions in the SDK.
```
ryan-rowland: How about a deprecated warning?

```
* @deprecated
* No longer used or updated. We now allow any string to be passed for region. Invalid regions
* won't be able to connect, and won't throw an exception.
```

lib/twilio/regions.ts
```diff
@@ -98,8 +105,84 @@ const regionURIs: Record<Region, string> = {
   [Region.Us2Tnx]: 'chunderw-vpc-gll-us2-tnx.twilio.com',
 };
 
+/**
+ * Check if a region is a long form AWS region
+ * @param region the region
+ * @returns a boolean representing whether or not the region is an AWS region
+ */
+const isAwsRegion = (region: string): boolean =>
+  (Object.keys(regionShortcodes) as string[]).includes(region);
```
ryan-rowland: What if amazon adds regions? Should we use a regex here instead?

lib/twilio/regions.ts
```diff
@@ -98,8 +105,84 @@ const regionURIs: Record<Region, string> = {
...
+ * @param region the region shortcode to transform
+ * @returns a string representing the chunder uri to perform a signaling
+ * connection to
+ */
+export const regionToChunderURI = (region: string | undefined): string => {
+  if (region === undefined || region === defaultRegion) {
+    return  `chunderw-vpc-gll.twilio.com`;
+  }
+
+  if (isAwsRegion(region)) {
```
ryan-rowland: Is this something we already supported or a new feature? If this is new, let's nix it imo. We'll have to maintain this list of known mappings.

lib/twilio/wstransport.ts
```diff
@@ -289,7 +289,19 @@ export default class WSTransport extends EventEmitter {
...
+  private _onSocketClose = (event: CloseEvent): void => {
+    this._log.info(`Received websocket close event code: ${event.code}`);
+    if (event.code === 1006) {
+      this.emit('error', {
+        code: 31005,
+        message: event.reason ||
+          'Websocket connection to Twilio\'s signaling servers were ' +
+          'unexpectedly ended. If this is happening consistently, there may ' +
+          'be an issue resolving the hostname provided. If a region is being ' +
+          'specified in Device setup, ensure it\'s a valid region.',
```
ryan-rowland: @aymenn Thoughts on this error text?

lib/twilio/wstransport.ts
```diff
@@ -289,7 +289,19 @@ export default class WSTransport extends EventEmitter {
   /**
    * Called in response to WebSocket#close event.
    */
-  private _onSocketClose = (): void => {
+  private _onSocketClose = (event: CloseEvent): void => {
+    this._log.info(`Received websocket close event code: ${event.code}`);
+    if (event.code === 1006) {
+      this.emit('error', {
```
ryan-rowland: Have we tested this is surfaced through Device? How does it look?

lib/twilio/regions.ts
mhuynh5757: This is something we already supported.

lib/twilio/regions.ts
```diff
@@ -98,31 +104,49 @@ const regionURIs: Record<Region, string> = {
...
+    return  `chunderw-vpc-gll.twilio.com`;
+  }
+
+  if (isRegionDeprecated(region)) {
     region = deprecatedRegions[region as DeprecatedRegion];
 
     // Don't let this callback affect script execution.
     if (onDeprecated) {
-      setTimeout(() => onDeprecated(region as Region));
+      setTimeout(() => onDeprecated(region!));
```
ryan-rowland: What's the exclamation point?

lib/twilio/regions.ts
```diff
@@ -98,31 +104,49 @@ const regionURIs: Record<Region, string> = {
   [Region.Us2Tnx]: 'chunderw-vpc-gll-us2-tnx.twilio.com',
 };
 
+/**
+ * Check whether or not the region is in the deprecated regions mapping.
+ * @param region the region
+ * @returns a boolean representing whether or not the region is deprecated
+ */
+export const isRegionDeprecated = (region: string): boolean =>
```
ryan-rowland: Doesn't look like we're using this outside of this file. We can probably move the logic into the one place we're consuming it.

CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+1.9.5 (In Progress)
+===================
+
+Improvements
+---------
+
+* In order to support new regions without requiring SDK updates in the future, the SDK no longer locally verifies the region option passed to `Device.setup` options. As a result, if an invalid region is passed, an exception won't be thrown and the Device will be unable to connect. We recommend always testing to make sure your specified region (if any) works before pushing any change to production. (CLIENT-6831)
```
aymenn: We could also add "For a list of supported regions refer to https://www.twilio.com/docs/voice/client/regions"

CHANGELOG.md
aymenn: Looks good small nit picks

To support future regions without requiring SDK updates,
   the SDK no longer verifies the region option passed to `Device.setup`
   options. As a result, if an invalid region is passed, a connection error will be emitted after the connection is attempted. We recommend always testing to make
   sure your specified region (if any) works before pushing any change to production.

What do you think?

lib/twilio/regions.ts
```diff
@@ -50,7 +56,7 @@ export type ValidRegion = Region | DeprecatedRegion;
 /**
  * Deprecated regions. Maps the deprecated region to its equivalent up-to-date region.
  */
-const deprecatedRegions: Record<DeprecatedRegion, Region> = {
+export const deprecatedRegions: Record<DeprecatedRegion, Region> = {
```
ryan-rowland: Why are we exporting this now? Necessary?

lib/twilio/regions.ts
```diff
@@ -98,31 +104,41 @@ const regionURIs: Record<Region, string> = {
   [Region.Us2Tnx]: 'chunderw-vpc-gll-us2-tnx.twilio.com',
 };
 
+/**
+ * The default region to connect to and create a chunder uri from if region is
+ * not defined.
+ * @constant string
+ */
+export const defaultRegion = 'gll';
```
ryan-rowland: Missing typecast, also not necessary to specify types in typedoc
```
/**
 * The default region to connect to and create a chunder uri from if region is
 * not defined.
 * @constant
 */
export const defaultRegion: string = 'gll';
```

lib/twilio/regions.ts
mhuynh5757: This is used for some tests in `tests/unit/regions.ts`.

tests/unit/device.ts
```diff
@@ -596,6 +596,13 @@ describe('Device', function() {
         clock.tick(30000 + 1);
         sinon.assert.notCalled(pstream.register);
       });
+
+      it('should emit Device.error if code is 31005', () => {
+        device.emit = sinon.spy();
+        pstream.emit('error', { error: { code: 31005 } });
+        sinon.assert.calledOnce(device.emit as any);
+        sinon.assert.calledWith(device.emit as any, 'error');
```
ryan-rowland: We should also test the format of the error payload emitted to make sure it's what we expect, in both cases (old error format, new error format)

lib/twilio/regions.ts
ryan-rowland: nvm I see it being used in tests.

PR #91, opened 2019-11-06
CLIENT-6897 | Dont hangup on getUserMedia failure

charliesantos:
**Contributing to Twilio**

### JIRA link(s):

- [CLIENT-6897](https://issues.corp.twilio.com/browse/CLIENT-6897)

### Description

Dont hangup on getUserMedia failure

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/lib/twilio/connection.ts
+++ b/lib/twilio/connection.ts
@@ -600,7 +600,9 @@ class Connection extends EventEmitter {
         }, this);
       }
 
-      this._disconnect();
+      // Reset the status back to pending
+      this._status = Connection.State.Pending;
+
       this.emit('error', { message, code });
     });
   }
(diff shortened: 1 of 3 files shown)


lib/twilio/connection.ts
```diff
@@ -566,6 +565,7 @@ class Connection extends EventEmitter {
       : this.mediaStream.openWithConstraints(audioConstraints);
 
     promise.then(() => {
+      this._status = Connection.State.Connecting;
```
ryan-rowland: This seems like a potentially API breaking change, since now the change is async rather than sync. Is this necessary for the fix?

charliesantos: It was needed so it can be retried after a failure. But I just realized this can be done better. I just need to set the status back to `pending` when getUserMedia fails. It now should not break the API and remains sync. Please see latest changes.

charliesantos:
Opened a new PR here https://github.com/twilio/twilio-client.js/pull/93/files
For some reason this is cached on travis and always fails when running npm commands.

PR #94, opened 2019-11-07
Client 6885 | ICE Gathering failures and timeouts issue ICE Restarts

charliesantos:
[CLIENT-6885](https://issues.corp.twilio.com/browse/CLIENT-6885) ICE Gathering failures and timeouts issue ICE Restarts.

### DIFF
--- a/lib/twilio/connection.ts
+++ b/lib/twilio/connection.ts
@@ -337,7 +337,7 @@ class Connection extends EventEmitter {
       this._publisher.post(level, 'ice-connection-state', state, null, this);
     };
 
-    this.mediaStream.onicegatheringfailure = (type: Connection.IceGatheringFailure): void => {
+    this.mediaStream.onicegatheringfailure = (type: Connection.IceGatheringFailureReason): void => {
       this._publisher.warn('ice-gathering-state', type, null, this);
       this._onMediaFailure(Connection.MediaFailure.IceGatheringFailed);
     };
@@ -1473,7 +1473,7 @@ namespace Connection {
   /**
    * Possible ICE Gathering failures
    */
-  export enum IceGatheringFailure {
+  export enum IceGatheringFailureReason {
     None = 'none',
     Timeout = 'timeout',
   }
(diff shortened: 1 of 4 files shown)


lib/twilio/rtc/peerconnection.js
```diff
@@ -654,11 +690,38 @@ PeerConnection.prototype._setupChannel = function() {
...
+      this._stopIceGatheringTimeout();
+
+      // Fail if no candidates found
+      if (!this._hasIceCandidates) {
+        this._onIceGatheringFailure(ICE_GATHERING_FAIL_TIMEOUT);
+      }
+
+      // There was a failure mid-gathering phase. We want to start our timer and issue
+      // an ice restart if we don't get connected after our timeout
+      if (this._hasIceCandidates && this._hasIceGatheringFailures) {
```
charliesantos: I'm not a fan of using more states in rtc/peerconnection, but I'm not sure how else we can capture these values without using boolean states. I'm open to suggestions!

lib/twilio/connection.ts
```diff
@@ -337,6 +337,11 @@ class Connection extends EventEmitter {
       this._publisher.post(level, 'ice-connection-state', state, null, this);
     };
 
+    this.mediaStream.onicegatheringfailure = (type: string): void => {
+      this._publisher.warn('ice-gathering-state', type, null, this);
```
ryan-rowland: What can this be, other than 'failed'?

ryan-rowland: Does this map to the following?
```
const ICE_GATHERING_FAIL_NONE = 'none';
const ICE_GATHERING_FAIL_TIMEOUT = 'timeout';
```

We should enumerate this, it's not clear what "type" can be.

Also, it looks like there's a case mistmatch here in `icegatheringfailure`

charliesantos: Ok, I will add an enum for 'none' and 'timeout'.

Are you referring to `_onIceGatheringFailure` and `onicegatheringfailure`

ryan-rowland: Nevermind on `onicegatheringfailure`, I see.

charliesantos: `_onIceGatheringFailure ` is a private method which calls `onicegatheringfailure ` which is a callback.  https://github.com/twilio/twilio-client.js/pull/94/files#diff-d6d1e696fa57bedaec729f3d6c8274a1R381

PR #96, opened 2019-11-08
CLIENT-6859 | Fixing CI script

charliesantos:
[CLIENT-6859](https://issues.corp.twilio.com/browse/CLIENT-6859) | Fixing CI script

### DIFF
--- a/.release.json
+++ b/.release.json
@@ -27,6 +27,8 @@
         "git push origin ${BRANCH} --tags",
         "git fetch --all",
         "git checkout ${RELEASE_VERSION}",
+        "git rm -rf docs",
+        "git commit -nm \"Preparing gh-pages\"",
         "git subtree add --prefix docs origin gh-pages",
         "npm run docs:ts",
         "git stage docs -f",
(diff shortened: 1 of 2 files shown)


.release.json
```diff
@@ -27,6 +27,8 @@
         "git push origin ${BRANCH} --tags",
         "git fetch --all",
         "git checkout ${RELEASE_VERSION}",
+        "git rm -rf docs",
+        "git commit -nm \"Preparing gh-pages\"",
```
charliesantos: We need to remove the docs folder from git first, or else the `git subtree` command will fail.

package.json
```diff
@@ -95,7 +95,7 @@
     "nyc": "10.1.2",
     "pre-commit": "1.2.2",
     "querystring": "0.2.0",
-    "release-tool": "0.2.2",
+    "release-tool": "git://github.com/twilio/release-tool#8860ca9",
```
charliesantos: Mak forked release tool to twilio org. I made some changes in the release-tool so that command line output are shown in the terminal which helps when debugging errors coming from release tool.

PR #100, opened 2019-11-18
CLIENT-6957 | Sending Local ICE candidates to insights

charliesantos:
**Contributing to Twilio**

### JIRA link(s):

- [CLIENT-6957](https://issues.corp.twilio.com/browse/CLIENT-6957)

### Description

Sending ICE candidates to insights.
`url` is not supported by the browsers. We cannot get stun/turn url for each candidate.

Only chrome supports these properties, as shown in the PR. Other browsers will only publish properties that are supported. See https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidate

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Improvements
 * Added additional triggers for media reconnection. ICE Restarts will be issued and will follow retry policy as defined in [media reconnection](https://www.twilio.com/docs/voice/client/javascript/connection#onreconnecting-handlererror) API.
   * ICE Gathering Failure - happens when ICE Gathering state transitions to `complete` and no ICE Candidates were successfully gathered.
   * ICE Gathering Timeout - happens when ICE Gathering exceeded 15 seconds and no ICE candidates were successfully gathered.
+* We now log ICE Candidates to insights during ICE gathering phase. (CLIENT-6957)
 
 1.9.5 (Nov 5, 2019)
 ===================
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ Improvements
 * Added additional triggers for media reconnection. ICE Restarts will be issued and will follow retry policy as defined in [media reconnection](https://www.twilio.com/docs/voice/client/javascript/connection#onreconnecting-handlererror) API.
   * ICE Gathering Failure - happens when ICE Gathering state transitions to `complete` and no ICE Candidates were successfully gathered.
   * ICE Gathering Timeout - happens when ICE Gathering exceeded 15 seconds and no ICE candidates were successfully gathered.
-* We now log ICE Candidates to insights during ICE gathering phase. (CLIENT-6957)
+* Locally gathered ICE candidates are now reported to Insights. (CLIENT-6957)
 
 1.9.5 (Nov 5, 2019)
 ===================
--- a/lib/twilio/connection.ts
+++ b/lib/twilio/connection.ts
@@ -334,7 +334,7 @@ class Connection extends EventEmitter {
     };
 
     this.mediaStream.onicecandidate = (candidate: IRTCIceCandidate): void => {
-      const payload = new RTCLocalIceCandidate(candidate).payload();
+      const payload = new RTCLocalIceCandidate(candidate).toPayload();
       this._publisher.debug('ice-candidate', 'ice-candidate', payload, this);
     };
 
--- a/lib/twilio/connection.ts
+++ b/lib/twilio/connection.ts
@@ -7,7 +7,7 @@ import { EventEmitter } from 'events';
 import Device from './device';
 import DialtonePlayer from './dialtonePlayer';
 import { GeneralErrors, InvalidArgumentError, MediaErrors, TwilioError } from './errors';
-import { IRTCIceCandidate, RTCLocalIceCandidate } from './rtc/candidate';
+import { RTCIceCandidate, RTCLocalIceCandidate } from './rtc/candidate';
 import RTCSample from './rtc/sample';
 import RTCWarning from './rtc/warning';
 import StatsMonitor from './statsMonitor';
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ Improvements
 * Added additional triggers for media reconnection. ICE Restarts will be issued and will follow retry policy as defined in [media reconnection](https://www.twilio.com/docs/voice/client/javascript/connection#onreconnecting-handlererror) API.
   * ICE Gathering Failure - happens when ICE Gathering state transitions to `complete` and no ICE Candidates were successfully gathered.
   * ICE Gathering Timeout - happens when ICE Gathering exceeded 15 seconds and no ICE candidates were successfully gathered.
+* We now log ICE Candidates to insights during ICE gathering phase. (CLIENT-6957)
```
idelgado: > Locally gathered ICE candidates are now reported to Insights.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ Improvements
...
```
idelgado: Please open a rtc-insights-spec PR to specify the JS version, modify the required fields of the existing JSON schema definition.

charliesantos: https://code.hq.twilio.com/client/rtc-insights-spec/pull/40

lib/twilio/connection.ts
```diff
@@ -1617,6 +1629,11 @@ namespace Connection {
      */
     getInputStream?: () => MediaStream;
 
+    /**
+     * A method to parse an ICE Candidate gathered by the browser and returns a RTCIceCandidate object
+     */
+    getRTCIceCandidate?: (candidate: BrowserIceCandidate) => RTCIceCandidate;
```
ryan-rowland: Why did we make this a public method, what would our customer use this for?

lib/twilio/rtc/candidate.ts
```diff
@@ -0,0 +1,90 @@
...
+    'protocol': ca.protocol,
+    'transport_id': ca.sdpMid,
+  };
+};
+
+/**
+ * Represents an ICE candidate coming from the browser
+ * @private
+ */
+export type BrowserIceCandidate = any;
```
ryan-rowland: Is this any or is this a string? If this is a string, can we nix `BrowserIceCandidate`?

lib/twilio/rtc/candidate.ts
```diff
@@ -0,0 +1,90 @@
...
+ * @module Voice
+ * @internalapi
+ */
+
+/**
+ * Parse an ICE candidate gathered by the browser and returns a RTCIceCandidate object
+ * @param candidate ICE candidate coming from the browser
+ * @returns RTCIceCandidate object
+ */
+export const getRTCIceCandidate = (ca: BrowserIceCandidate): RTCIceCandidate => {
```
ryan-rowland: Can we merge `getRTCIceCandidate` and `RTCIceCandidate` into an `RTCIceCandidate` class? The constructor can take a string (the BrowserIceCandidate).

lib/twilio/connection.ts
charliesantos: This is adding to connection.options, so we can pass it in. It's private https://github.com/twilio/twilio-client.js/pull/100/files#diff-2ec8416c9cc5a81e6e7db4b6d8966f51R148. But shouldn't matter because we're converting this into a class

lib/twilio/rtc/candidate.ts
charliesantos: It's an object. Most of the properties we need we grab from this object directly, but some properties, such as network-cost, is parsed from the 'candidate' property of this object.

lib/twilio/connection.ts
ryan-rowland: Ah hah

lib/twilio/rtc/candidate.ts
```diff
@@ -0,0 +1,109 @@
...
+    this.port = iceCandidate.port;
+    this.priority = iceCandidate.priority;
+    this.protocol = iceCandidate.protocol;
+    this.transportId = iceCandidate.sdpMid;
+  }
+
+  /**
+   * Get the payload object for insights
+   */
+  payload(): IRTCIceCandidateInsights {
```
ryan-rowland: How about `toPayload(): RTCIceCandidatePayload`?

lib/twilio/rtc/candidate.ts
```diff
@@ -0,0 +1,109 @@
+/**
+ * @module Voice
+ * @internalapi
+ */
+
+/**
+ * Payload bject we send to insights
```
ryan-rowland: sp: object

lib/twilio/rtc/candidate.ts
```diff
@@ -0,0 +1,109 @@
...
+ * @private
+ */
+type RTCIceCandidatePayload = any;
```
ryan-rowland: Should we define this, actually?

lib/twilio/rtc/candidate.ts
```diff
@@ -0,0 +1,109 @@
...
+
+/**
+ * Represents an ICE candidate coming from the browser
+ * https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidate
+ * @private
+ */
+export type IRTCIceCandidate = any;
```
ryan-rowland: Do we use the `I*` interface naming scheme? Also, since this is called `RTCIceCandidate` maybe that's what we should name it?

charliesantos: We use `I*` on other files. Should we start not using the `I*` scheme?

lib/twilio/rtc/candidate.ts
charliesantos: I've added an interface. I did not add ts docs because it's the exact copy of the docs in the `RTCLocalIceCandidate` class below. Let me know what you think.

PR #104, opened 2019-11-22
CLIENT-6970 | Running preflight test

charliesantos:
For discussion only. Some things I came across while implementing the spec;

- Namespace Voice is already a module
- need a "To" field number or identity. Should we parameterize it?
- Regarding "_returns a rejection with a VoiceDiagnostics.FatalError when encountering a fatal error_"
  -  should we just return the Preflight object? Since we already have on('failed')
- CodecName type is already in connection class. Why need a new type?
- When preflight.cancel is called, should we call onComplete or onFailed?
- for on('warning'), do we really want to emit RTCWarning, or just the public warning events?
- we need to list more errors: Fatal and NonFatal
- exposing eventpublisher errors, went for eventemitter, let's rethink this


lib/twilio/connection.ts
```diff
@@ -287,6 +287,10 @@ class Connection extends EventEmitter {
         this._onMediaFailure(Connection.MediaFailure.LowBytes);
       }
       this._reemitWarning(data, wasCleared);
+
+      // TODO: Re-think this
+      // For getting RTCWarning directly
+      this.emit('warning', data);
```
ryan-rowland: What's this? We can't change the public API this way

lib/twilio/device.ts
```diff
@@ -24,6 +24,7 @@ import {
 } from './regions';
 import Log, { LogLevel } from './tslog';
 import { Exception, queryToJson } from './util';
+import Voice from './voice';
```
ryan-rowland: We call this `Voice` in the spec, I know, but it's too ambiguous here. Let's call this something else like:
```
import { PreflightTest } from './preflight;
```

lib/twilio/eventpublisher.js
```diff
@@ -109,8 +113,12 @@ EventPublisher.prototype._post = function _post(endpointName, level, group, name
   const self = this;
   return new Promise((resolve, reject) => {
     self._request.post(requestParams, err => {
-      if (err) { reject(err); }
-      else { resolve(); }
+      if (err) {
+        this.emit('error', err);
```
ryan-rowland: Can you explain why this is necessary? I think we should be able to catch errors in the returned Promise, and not need to add an error event that fires at the same time.

lib/twilio/statsMonitor.ts
```diff
@@ -305,9 +305,11 @@ class StatsMonitor extends EventEmitter {
     const rttValue = (typeof stats.rtt === 'number' || !previousSample) ? stats.rtt : previousSample.rtt;
 
     return {
-      audioInputLevel: Math.round(average(this._inputVolumes.splice(0))),
+      // TODO: Undo this. For testing only
+      audioInputLevel: 0,
       audioOutputLevel: Math.round(average(this._outputVolumes.splice(0))),
-      bytesReceived: currentBytesReceived,
+      // TODO: Undo this. For testing only
+      bytesReceived: 0,
```
ryan-rowland: Don't forget these testing values :)

lib/twilio/voice.ts
```diff
@@ -0,0 +1,176 @@
...
+ * @preferred
+ * @publicapi
+ */
+import { EventEmitter } from 'events';
+import Connection from './connection';
+import Device from './device';
+import RTCSample from './rtc/sample';
+import RTCWarning from './rtc/warning';
+
+ namespace Voice {
```
ryan-rowland: Yeah we can drop this namespace, this is more for the Diagnostic SDK. We can just return the classes from here instead.

lib/twilio/voice.ts
```diff
@@ -0,0 +1,176 @@
...
+  export class PreflightTest extends EventEmitter {
+
+    private _connection: Connection;
+    private _device: Device;
+    private _errors: NonFatalError[];
+    private _samples: RTCSample[];
+    private _warnings: RTCWarning[];
+
+    private _startTime: number | undefined;
+    private _endTime: number | undefined;
```
ryan-rowland: Alphabetical? 😛

lib/twilio/voice.ts
```diff
@@ -0,0 +1,176 @@
...
+    private _latestSample: RTCSample | undefined;
+    private _results: TestResults | undefined;
+    private _status: TestStatus = TestStatus.Connecting;
+
+    constructor(token: string, options: PreflightOptions) {
+      super();
+
+      const codecPreferences = options.codecPreferences && options.codecPreferences.length ?
+        options.codecPreferences : [Connection.Codec.Opus, Connection.Codec.PCMU];
```
ryan-rowland: Can we set a default options object instead? See Device and Connection for how we do this.

lib/twilio/voice.ts
```diff
@@ -0,0 +1,176 @@
...
+      // For getting RTCWarning directly
+      conn.on('warning', (warning: RTCWarning) => {
+        if (typeof warning !== 'string' && warning) {
+          this._warnings.push(warning);
+          this.emit('warning', warning);
+        }
+      });
+    }
+
+    get startTime(): number | undefined {
```
ryan-rowland: This should always be defined, right? Start time should be construction time, I think

lib/twilio/connection.ts
charliesantos: On spec, we want to get RTCWarning data. But connection doesn't currently emit this data directly, just the name of the warning. I need to rethink this.

ryan-rowland:
> Namespace Voice is already a module

Yep, this was more for the Diagnostic SDK. Let's nix it.

> need a "To" field number or identity. Should we parameterize it?

Yeah, this is hard. We need to think on it more. Twilio should have a global template a customer can use to make an outgoing call without needing to set up a TwiML app.

> Regarding "returns a rejection with a VoiceDiagnostics.FatalError when encountering a fatal error"
> should we just return the Preflight object? Since we already have on('failed')

Yes, you're right. We're not returning a Promise so let's just surface this through on('failed') and represent the fatal error in the report.

> CodecName type is already in connection class. Why need a new type?

We can use the existing CodecName

> When preflight.cancel is called, should we call onComplete or onFailed?

Hmm, maybe onFailed? We'd need to have a reason that represents cancellation though.

> for on('warning'), do we really want to emit RTCWarning, or just the public warning events?

The customer shouldn't have access to the Device being used for these tests so those warnings should be fine to emit, right? We probably want to disable Insights for the whole test call though, and only send the resulting report to Insights, including the CallSid.

> we need to list more errors: Fatal and NonFatal

If there are any missing, let's add them

> exposing eventpublisher errors, went for eventemitter, let's rethink this

I agree

lib/twilio/eventpublisher.js
charliesantos: Either this or we would need to add catch blocks for all publisher.post/warn/debug and emit the event as a new error, maybe an insights failed error. Still thinking about these solutions.

ryan-rowland: Hmm, this might make sense. Let's think on it some more.

charliesantos: I thought through it and it seemed to make sense to emit this error, similar pattern we're using on most of our files.

charliesantos:
@ryan-rowland This is ready for another review. I've addressed your previous comments and completed implementation. If you like the direction, I will write tests. The following are new that are not captured in the spec doc. We'll review with Aymen later.
- VoiceJS SDK now publishes insights endpoint errors. We are also using this in the preflight to capture non-fatal errors
- VoiceJS SDK now publishes RTCWarning data
- Parameterized preflightTest's connect params. These are the parameters passed to TwiML

lib/twilio/preflight.ts
```diff
@@ -0,0 +1,409 @@
...
+
+    /**
+     * Parameters that will be sent to your Twilio Application via {@link Device.connect}
+     */
+    connectParams: Record<string, string>;
+
+    /**
+     * Device class to use
+     */
+    deviceFactory?: new (token: string, options: Device.Options) => Device;
```
ryan-rowland: We should split out our testing options and the public facing options. See DeviceOptions and ExtendedDeviceOptions.

lib/twilio/preflight.ts
```diff
@@ -0,0 +1,409 @@
...
+  }
+
+  /**
+   * Represents the warning emitted from Voice SDK
+   */
+  export interface PreflightWarning {
```
ryan-rowland: Any reason we can't just expose RTCWarning here?

lib/twilio/device.ts
```diff
@@ -1399,6 +1439,37 @@ namespace Device {
    */
   export type ToggleableSound = Device.SoundName.Incoming | Device.SoundName.Outgoing | Device.SoundName.Disconnect;
 
+  /**
+   * The error format used by errors emitted from {@link Device}.
+   */
+  export interface Error {
```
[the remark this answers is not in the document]
charliesantos: No, we don't have it defined. We are emitting Device error objects without at type.
Right now, we reference it as `Device.Error`. I can rename this interface to `DeviceError` but it'd be weird to reference it to other files as `Device.DeviceError`. Thoughts?

lib/twilio/preflight.ts
```diff
@@ -0,0 +1,409 @@
...
+   * @param connectParams - Parameters that will be sent to your Twilio Application via {@link Device.connect}
+   */
+  private _onDeviceReady(connectParams: Record<string, string>): void {
+    this._connection = this._device.connect(connectParams);
+    this._setupConnectionHandlers(this._connection);
+
+    this._callTimeoutId = setTimeout(() => {
+      this._device.once('offline', () => this._onCompleted());
+      this._device.destroy();
+    }, (this._options.callSeconds! * 1000));
```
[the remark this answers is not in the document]
charliesantos: `! post-fix operator` - meaning, this cannot be null or undefined in this situation.

`callSeconds` is an optional parameter. Without `!` operator, typescript complains that this might be undefined. However, we are setting a default in this class, but the compiler cannot detect that there's a default. This is a recommended way I've seen to address that.

lib/twilio/preflight.ts
charliesantos: We can, but `PreflightWarning` contains not just `RTCWarning`. It also contains what type of warning is emitted such as `constant-audio`, `low-rtt`, `high-foo` etc. `Connection` is already doing this for us. We are basically just re-emitting what we receive from connection.

lib/twilio/preflight.ts
```diff
@@ -0,0 +1,409 @@
...
+      return totals;
+    }, {});
+
+    // Make a copy of a sample, along with the totals
+    const sample = Object.assign({}, this._latestSample, sampleTotals);
+
+    // Calculate the average
+    keys.forEach((key: string) => {
+      sample[key] = sample[key] / this._samples.length;
+    });
```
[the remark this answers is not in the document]
charliesantos: Hmm, thinking more about this, I don't think `.map` will work here. We are not trying to transform the elements.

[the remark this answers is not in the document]
ryan-rowland: How about a .reduce?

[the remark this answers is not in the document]
charliesantos: Reduce is already used at the top to get totals. 2nd step is getting the average. I thought a 2-step process would be easier to read. I now updated this whole function to use cumulative averaging so it's only a one step process. Let me know if you like this better. See latest commit here https://github.com/twilio/twilio-client.js/pull/104/commits/e91e7d6aa2efc848da80e91c813ea12f909fa56d

PR #106, opened 2019-11-25
CLIENT-7025 | Parameterize ICE Aggressive Nomination

charliesantos:
[CLIENT-7025](https://issues.corp.twilio.com/browse/CLIENT-7025)

### DIFF
--- a/lib/twilio/connection.ts
+++ b/lib/twilio/connection.ts
@@ -297,8 +297,8 @@ class Connection extends EventEmitter {
         codecPreferences: this.options.codecPreferences,
         debug: this.options.debug,
         dscp: this.options.dscp,
-        enableIceAggressiveNomination: this.options.enableIceAggressiveNomination,
         enableIceRestart: this.options.enableIceRestart,
+        forceAggressiveIceNomination: this.options.forceAggressiveIceNomination,
         isUnifiedPlan: this._isUnifiedPlanDefault,
         maxAverageBitrate: this.options.maxAverageBitrate,
         warnings: this.options.warnings,
@@ -1603,12 +1603,6 @@ namespace Connection {
     dscp?: boolean;
 
     /**
-     * Experimental feature.
-     * Whether to use ICE Aggressive nomination.
-     */
-    enableIceAggressiveNomination?: boolean;
-
-    /**
      * Whether to automatically restart ICE when media connection fails
      */
     enableIceRestart?: boolean;
@@ -1617,6 +1611,12 @@ namespace Connection {
      * Whether the ringing state should be enabled.
      */
     enableRingingState?: boolean;
+
+    /**
+     * Experimental feature.
+     * Whether to use ICE Aggressive nomination.
+     */
+    forceAggressiveIceNomination?: boolean;
 
     /**
      * The gateway currently connected to.
--- a/lib/twilio/connection.ts
+++ b/lib/twilio/connection.ts
@@ -1614,7 +1614,7 @@ namespace Connection {
 
     /**
      * Experimental feature.
-     * Whether to use ICE Aggressive nomination.
+     * Force Chrome's ICE agent to use aggressive nomination when selecting a candidate pair.
      */
     forceAggressiveIceNomination?: boolean;
 
--- a/lib/twilio/connection.ts
+++ b/lib/twilio/connection.ts
@@ -297,6 +297,7 @@ class Connection extends EventEmitter {
         codecPreferences: this.options.codecPreferences,
         debug: this.options.debug,
         dscp: this.options.dscp,
+        enableIceAggressiveNomination: this.options.enableIceAggressiveNomination,
         enableIceRestart: this.options.enableIceRestart,
         isUnifiedPlan: this._isUnifiedPlanDefault,
         maxAverageBitrate: this.options.maxAverageBitrate,
@@ -1600,6 +1601,12 @@ namespace Connection {
      * Whether or not to enable DSCP.
      */
     dscp?: boolean;
+
+    /**
+     * Experimental feature.
+     * Whether to use ICE Aggressive nomination.
+     */
+    enableIceAggressiveNomination?: boolean;
 
     /**
      * Whether to automatically restart ICE when media connection fails
--- a/lib/twilio/rtc/peerconnection.js
+++ b/lib/twilio/rtc/peerconnection.js
@@ -7,6 +7,7 @@ const {
 const Log = require('../log');
 const util = require('../util');
 const RTCPC = require('./rtcpc');
+const { setIceAggressiveNomination } = require('./sdp');
 
 const ICE_GATHERING_TIMEOUT = 15000;
 const ICE_GATHERING_FAIL_NONE = 'none';
@@ -652,6 +653,11 @@ PeerConnection.prototype._setupPeerConnection = function(rtcConstraints, rtcConf
   };
   return version;
 };
+
+PeerConnection.prototype._maybeSetIceAggressiveNomination = function(sdp) {
+  return this.options.enableIceAggressiveNomination ? setIceAggressiveNomination(sdp) : sdp;
+};
+
 PeerConnection.prototype._setupChannel = function() {
   const pc = this.version.pc;
 
@@ -836,7 +842,12 @@ PeerConnection.prototype.makeOutgoingCall = function(token, params, callsid, rtc
 
     self._answerSdp = payload.sdp;
     if (self.status !== 'closed') {
-      self.version.processAnswer(this.codecPreferences, payload.sdp, onAnswerSuccess, onAnswerError);
+      self.version.processAnswer(
+        this.codecPreferences,
+        this._maybeSetIceAggressiveNomination(payload.sdp),
+        onAnswerSuccess,
+        onAnswerError
+      );
     }
     self.pstream.removeListener('answer', self._onAnswerOrRinging);
     self.pstream.removeListener('ringing', self._onAnswerOrRinging);
@@ -892,7 +903,14 @@ PeerConnection.prototype.answerIncomingCall = function(callSid, sdp, rtcConstrai
       twilioError: new MediaErrors.ClientRemoteDescFailed(),
     } });
   }
-  this.version.processSDP(this.options.maxAverageBitrate, this.codecPreferences, sdp, { audio: true }, onAnswerSuccess, onAnswerError);
+  this.version.processSDP(
+    this.options.maxAverageBitrate,
+    this.codecPreferences,
+    this._maybeSetIceAggressiveNomination(sdp),
+    { audio: true },
+    onAnswerSuccess,
+    onAnswerError
+  );
 };
 PeerConnection.prototype.close = function() {
   if (this.version && this.version.pc) {
--- a/lib/twilio/rtc/peerconnection.js
+++ b/lib/twilio/rtc/peerconnection.js
@@ -789,9 +789,10 @@ PeerConnection.prototype.iceRestart = function() {
         return;
       }
 
-      this._answerSdp = payload.sdp;
+      const sdp = this._maybeSetIceAggressiveNomination(payload.sdp);
+      this._answerSdp = sdp;
       if (this.status !== 'closed') {
-        this.version.processAnswer(this.codecPreferences, payload.sdp, null, err => {
+        this.version.processAnswer(this.codecPreferences, sdp, null, err => {
           const message = err && err.message ? err.message : err;
           this.log(`Failed to process answer during ICE Restart. Error: ${message}`);
         });
@@ -880,6 +881,7 @@ PeerConnection.prototype.answerIncomingCall = function(callSid, sdp, rtcConstrai
   if (!this._initializeMediaStream(rtcConstraints, rtcConfiguration)) {
     return;
   }
+  sdp = this._maybeSetIceAggressiveNomination(sdp);
   this._answerSdp = sdp.replace(/^a=setup:actpass$/gm, 'a=setup:passive');
   this.callSid = callSid;
   const self = this;
@@ -903,14 +905,7 @@ PeerConnection.prototype.answerIncomingCall = function(callSid, sdp, rtcConstrai
       twilioError: new MediaErrors.ClientRemoteDescFailed(),
     } });
   }
-  this.version.processSDP(
-    this.options.maxAverageBitrate,
-    this.codecPreferences,
-    this._maybeSetIceAggressiveNomination(sdp),
-    { audio: true },
-    onAnswerSuccess,
-    onAnswerError
-  );
+  this.version.processSDP(this.options.maxAverageBitrate, this.codecPreferences, sdp, { audio: true }, onAnswerSuccess, onAnswerError);
 };
 PeerConnection.prototype.close = function() {
   if (this.version && this.version.pc) {
--- a/lib/twilio/rtc/peerconnection.js
+++ b/lib/twilio/rtc/peerconnection.js
@@ -841,14 +841,10 @@ PeerConnection.prototype.makeOutgoingCall = function(token, params, callsid, rtc
   this._onAnswerOrRinging = payload => {
     if (!payload.sdp) { return; }
 
-    self._answerSdp = payload.sdp;
+    const sdp = this._maybeSetIceAggressiveNomination(payload.sdp);
+    self._answerSdp = sdp;
     if (self.status !== 'closed') {
-      self.version.processAnswer(
-        this.codecPreferences,
-        this._maybeSetIceAggressiveNomination(payload.sdp),
-        onAnswerSuccess,
-        onAnswerError
-      );
+      self.version.processAnswer(this.codecPreferences, sdp, onAnswerSuccess, onAnswerError);
     }
     self.pstream.removeListener('answer', self._onAnswerOrRinging);
     self.pstream.removeListener('ringing', self._onAnswerOrRinging);
--- a/lib/twilio/rtc/peerconnection.js
+++ b/lib/twilio/rtc/peerconnection.js
@@ -655,7 +655,7 @@ PeerConnection.prototype._setupPeerConnection = function(rtcConstraints, rtcConf
 };
 
 PeerConnection.prototype._maybeSetIceAggressiveNomination = function(sdp) {
-  return this.options.enableIceAggressiveNomination ? setIceAggressiveNomination(sdp) : sdp;
+  return this.options.forceAggressiveIceNomination ? setIceAggressiveNomination(sdp) : sdp;
 };
 
 PeerConnection.prototype._setupChannel = function() {
--- a/lib/twilio/rtc/sdp.js
+++ b/lib/twilio/rtc/sdp.js
@@ -18,6 +18,8 @@ function getPreferredCodecInfo(sdp) {
 
 function setIceAggressiveNomination(sdp) {
   // This only works on Chrome. We don't want any side effects on other browsers
+  // https://bugs.chromium.org/p/chromium/issues/detail?id=1024096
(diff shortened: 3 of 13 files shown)


lib/twilio/rtc/sdp.js
```diff
@@ -16,6 +16,17 @@ function getPreferredCodecInfo(sdp) {
   return { codecName, codecParams };
 }
 
+function setIceAggressiveNomination(sdp) {
+  // This only works on Chrome. We don't want any side effects on other browsers
+  if (!util.isChrome(window, window.navigator)) {
```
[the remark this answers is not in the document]
charliesantos: There's currently no way to detect this functionality on the browsers. Unfortunately, aggressive nomination is not something that's exposed by browser. We also cannot inspect the SDP format to determine this. This sdp is coming from the server and it's the same for all ice-agents.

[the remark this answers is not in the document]
ceaglest: The issue is somewhat specific to Chrome's ICE agent, so I don't think inspecting the SDP will give us enough information.

lib/twilio/rtc/peerconnection.js
```diff
@@ -836,7 +842,12 @@ PeerConnection.prototype.makeOutgoingCall = function(token, params, callsid, rtc
 
     self._answerSdp = payload.sdp;
     if (self.status !== 'closed') {
-      self.version.processAnswer(this.codecPreferences, payload.sdp, onAnswerSuccess, onAnswerError);
+      self.version.processAnswer(
+        this.codecPreferences,
+        this._maybeSetIceAggressiveNomination(payload.sdp),
```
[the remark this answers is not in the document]
charliesantos: It's only used on plan-b. Is that still applicable for this situation?

lib/twilio/connection.ts
```diff
@@ -1601,6 +1602,12 @@ namespace Connection {
      */
     dscp?: boolean;
 
+    /**
+     * Experimental feature.
+     * Whether to use ICE Aggressive nomination.
+     */
+    enableIceAggressiveNomination?: boolean;
```
ceaglest: Is there an IDL we could add this to for Voice? I guess for Video I would consider adding it to our existing ConnectOptions / IceOptions API. 

Naming suggestion, since this flag is describing a workaround that might stop being effective in the future:

```suggestion
    forceAggressiveIceNomination?: boolean;
```

lib/twilio/connection.ts
```diff
@@ -1601,6 +1602,12 @@ namespace Connection {
      */
     dscp?: boolean;
 
+    /**
+     * Experimental feature.
+     * Whether to use ICE Aggressive nomination.
```
ceaglest:
```suggestion
     * Force Chrome's ICE agent to use aggressive nomination when selecting a candidate pair.
```

lib/twilio/rtc/sdp.js
```diff
@@ -16,6 +16,17 @@ function getPreferredCodecInfo(sdp) {
   return { codecName, codecParams };
 }
 
+function setIceAggressiveNomination(sdp) {
+  // This only works on Chrome. We don't want any side effects on other browsers
```
ceaglest: Link the chromium bug, and maybe our internal ticket for context.

What about other libWebRTC ICE agents, like Safari or Opera? If those browsers have the same problem we **might** want to target them as well. @syerrapragada might have thoughts on this.

lib/twilio/rtc/peerconnection.js
```diff
@@ -892,7 +903,14 @@ PeerConnection.prototype.answerIncomingCall = function(callSid, sdp, rtcConstrai
...
   }
-  this.version.processSDP(this.options.maxAverageBitrate, this.codecPreferences, sdp, { audio: true }, onAnswerSuccess, onAnswerError);
+  this.version.processSDP(
+    this.options.maxAverageBitrate,
+    this.codecPreferences,
+    this._maybeSetIceAggressiveNomination(sdp),
+    { audio: true },
+    onAnswerSuccess,
+    onAnswerError
+  );
```
ceaglest: Are you forcing the SDK to be the DTLS Client by munging the incoming SDP on line 883? Is this needed for specific browsers?

I would rather rewrite our local SDP before applying it if DTLS role switching is needed for Chrome, FF or others.

tests/peerconnection.js
```diff
@@ -889,6 +899,14 @@ describe('PeerConnection', () => {
       assert(context.pstream.on.calledWithExactly('answer', sinon.match.func));
     });
 
+    it('Should enable ice aggressive nomination when status not closed and answer is emitted', () => {
+      context.status = NOT_CLOSED;
+      context.pstream.on.callsArgWith(1, PAYLOAD);
+      context._maybeSetIceAggressiveNomination = sinon.stub();
+      toTest();
+      sinon.assert.calledWithExactly(context._maybeSetIceAggressiveNomination, PAYLOAD.sdp);
+    });
```
ceaglest: This is for inbound calls, and above is for outbound?

tests/peerconnection.js
```diff
@@ -939,6 +957,35 @@ describe('PeerConnection', () => {
     });
   });
 
+  context('PeerConnection.prototype._maybeSetIceAggressiveNomination', () => {
+    const METHOD = PeerConnection.prototype._maybeSetIceAggressiveNomination;
+    const USER_AGENT = root.window.navigator.userAgent;
+    const SDP = 'a=ice-lite\nfoo';
```
ceaglest:
```suggestion
    const SDP = 'bar\na=ice-lite\nfoo';
```

tests/sdp.js
```diff
@@ -2,11 +2,43 @@
...
+  setIceAggressiveNomination,
+  setMaxAverageBitrate
+} = require('../lib/twilio/rtc/sdp');
 
 const { makeSdpWithTracks } = require('./lib/mocksdp');
 const { combinationContext } = require('./lib/util');
 
+describe('setIceAggressiveNomination', () => {
+  const SDP_ICE_LITE = 'a=ice-lite\nfoo';
+  const SDP_FULL_ICE = 'a=group\nfoo';
```
ceaglest: Make it more representative of the real thing?

```suggestion
  const SDP_FULL_ICE = 'a=group\nfoo\na=ice-options:trickle-ice\n';
```

lib/twilio/rtc/sdp.js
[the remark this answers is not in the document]
ceaglest: Walking back on this statement, I think other browsers (Safari, Opera, Chromium based) would be impacted as well. I'm not sure how we can detect that they are using the libWebRTC ICE agent, but if we could that would be the best way.

lib/twilio/rtc/peerconnection.js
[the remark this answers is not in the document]
ceaglest: What if there is an ice restart?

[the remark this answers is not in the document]
charliesantos: ICE restart won't have this. I intentionally added this only at the initial call setup to minimize the risk/potential problems. When ice restart happens, there would still be a delay but I don't think it would be a UX issue. I we still think we should add it on ice restart, we need to make sure this is covered/tested.

lib/twilio/rtc/sdp.js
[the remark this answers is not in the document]
charliesantos: I'm not sure if we want to cover other browsers for this round. We should discuss this with the team/Aymen.

lib/twilio/connection.ts
charliesantos: Sounds good on the name suggestion. I don't think we have IDL for voice js :(

lib/twilio/rtc/sdp.js
charliesantos: I suggest we should limit our scope for this round since we have a customer that really needs this soon. But yeah, let's discuss with the team including Aymen.

tests/peerconnection.js
charliesantos: Line 902 is for outgoing. Line 511 is for incoming

lib/twilio/rtc/peerconnection.js
charliesantos: @ryan-rowland what was the reason for line 883?

ryan-rowland: I believe this has been in effect since before I worked on client :) I am not entirely sure of the reasoning behind it.

charliesantos: Sounds like this should be on a different discussion and shouldn't be a part of this PR. I don't want to change something that works if we don't fully understand it yet.

lib/twilio/rtc/sdp.js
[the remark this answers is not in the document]
ceaglest: > We should discuss this with the team/Aymen.

Yes I agree.

We should focus on adding test coverage to characterize the problem in each browser that we support. This PR is entirely about SDP processing, and validating that the processing is working correctly. This isn't actually telling us if the problem occurs or not on each browser.

lib/twilio/rtc/peerconnection.js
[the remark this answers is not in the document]
ceaglest: It seems very strange to change nomination strategies mid-call. I would suggest that we add the test coverage that is needed rather than using two different nomination strategies.

lib/twilio/rtc/peerconnection.js
ceaglest: I am trying to understand why we need to enforce role switching on the Client when this could be done by the Server. I agree that it is out of scope.

charliesantos: Let's also keep in mind, this role switching is only used for plan-b. And just like Ryan, I'm not entirely sure why it was added.

lib/twilio/rtc/sdp.js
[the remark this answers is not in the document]
charliesantos:
```
BROWSER       Delay Issue    Fix works
---------------------------------------
Chrome          yes           yes
Chromium Edge   yes           yes
Safari          yes           no
Firefox         no
Opera (not supported)
```
Considering the results, I think we should only target this fix for Chrome and Chromium Edge

lib/twilio/rtc/peerconnection.js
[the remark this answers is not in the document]
charliesantos: I think you're right. I updated it so ice restart have this too.

tests/integration/icenomination.ts
```diff
@@ -0,0 +1,138 @@
...
+        deviceOptions = direction === Connection.CallDirection.Outgoing ? device1Options : device2Options;
+      });
+  
+      afterEach(() => {
+        device1Options = {};
+        device2Options = {};
+        destroyDevices();
+      });
+  
+      it('should not have dtls handshake delay if region is us1 and aggressive nomination is false', async () => {
```
ceaglest: This statement, and the test results are inaccurate. With Chrome and TMS there is always handshake delay, there are just 3 buckets of increasing delay. If the intention is to evaluate the "short" delay then maybe you could determine it in terms of a fraction of a RTT.

tests/integration/icenomination.ts
```diff
@@ -0,0 +1,138 @@
...
+    // We don't currently expose icegathering state changes.
+    // Let's hook to makeConnection and subscribe to events
+    const makeConnection = device['_makeConnection'].bind(device);
+    (device['_makeConnection'] as any) = (...params: any) => {
+      const conn = makeConnection(...params);
+      conn.mediaStream.onicegatheringstatechange = (state: string) => {
+        if (state === 'gathering') {
+          start = Date.now();
+        }
+      };
```
ceaglest: Do we care about gathering the local candidates? If it takes a long time to bind a socket, I don't see how that is related to this bug.

I think we should consider measuring: 

`oniceconnectionstatechange` : 'checking' to 
`oniceconnectionstatechange` 'connected'

In order to estimate RTT and determine if we are testing a short, medium or long delay scenario or not. And then finally:

`oniceconnectionstatechange` : 'checking' to 
`onpcconnectionstatechange ` : 'connected'

To determine the time taken for ICE and DTLS to complete.

tests/integration/icenomination.ts
```diff
@@ -0,0 +1,138 @@
...
+        destroyDevices();
+      });
+  
+      it('should not have dtls handshake delay if region is us1 and aggressive nomination is false', async () => {
+        deviceOptions.region = 'us1';
+        await setupDevices();
+        await getConnectionDuration(direction).then(duration => {
+          assert(duration < CONNECTION_DELAY_THRESHOLD);
+        });
+      });
```
ceaglest: This test is specific to a particular region, and assumes that the device under test is also near us1 and has 1 network interface (I think? or else you would not reproduce the longer delay with sg1..).

Do you have a guarantee on where your CI testers are running, and that they have < 96 ms RTT to us1? Should the test be skipped, or the results discounted if the opening RTT is longer than this?

tests/integration/icenomination.ts
```diff
@@ -0,0 +1,138 @@
...
+  
+      it('should not have dtls handshake delay if region is us1 and aggressive nomination is true', async () => {
+        deviceOptions.region = 'us1';
+        deviceOptions.forceAggressiveIceNomination = true;
+        await setupDevices();
+        await getConnectionDuration(direction).then(duration => {
+          assert(duration < CONNECTION_DELAY_THRESHOLD);
+        });
+      });
```
ceaglest: Is it possible to prove that the handshake completes more quickly when aggressive nomination is true? This test and the previous one use the same duration threshold.

tests/integration/icenomination.ts
```diff
@@ -0,0 +1,138 @@
...
+  
+      it('should have dtls handshake delay if region is sg1 and aggressive nomination is false', async () => {
+        deviceOptions.region = 'sg1';
+        await setupDevices();
+        await getConnectionDuration(direction).then(duration => {
+          assert(duration > CONNECTION_DELAY_THRESHOLD);
+        });
+      });
```
ceaglest: I'm not sure how you can generally make this assertion unless you know exactly where your tester will be when the test runs, and if its RTT < 96 ms.

tests/unit/device.ts
```diff
@@ -103,6 +103,21 @@ describe('Device', function() {
...
+      const conn = device.setup(token, setupOptions);
+      assert.equal(conn['options'].forceAggressiveIceNomination, false);
+    });
+
+    it('should set forceAggressiveIceNomination to false if passed in as false', () => {
+      const conn = device.setup(token, Object.assign({ forceAggressiveIceNomination: false}, setupOptions));
+      assert.equal(conn['options'].forceAggressiveIceNomination, false);
+    });
+
+    it('should set forceAggressiveIceNomination to false if passed in as true', () => {
```
ceaglest: Possible typo or my misunderstand of JavaScript programming? 

```suggestion
    it('should set forceAggressiveIceNomination to true if passed in as true', () => {
```

tests/integration/icenomination.ts
```diff
@@ -0,0 +1,138 @@
...
+import { expectEvent, isFirefox } from '../lib/util';
+import * as assert from 'assert';
+
+const CONNECTION_DELAY_THRESHOLD = 1000;
+const SUITE_TIMEOUT = 20000;
+const MAX_TIMEOUT = 300000;
+
+const maybeSkip = isFirefox() ? describe.skip : describe;
+
+maybeSkip('ICE Nomination', function() {
```
ceaglest: Should be skipping some of the tests on Firefox, or expecting a different outcome? It might still be useful to test in case Firefox decides to fix their bugs and implement standards compliant behavior.

tests/integration/icenomination.ts
```diff
@@ -0,0 +1,138 @@
...
+      });
+  
+      it('should not have dtls handshake delay if region is sg1 and aggressive nomination is true', async () => {
+        deviceOptions.region = 'sg1';
+        deviceOptions.forceAggressiveIceNomination = true;
+        await setupDevices();
+        await getConnectionDuration(direction).then(duration => {
+          assert(duration < CONNECTION_DELAY_THRESHOLD);
+        });
+      });
```
ceaglest: Why `sg1` and not any other Twilio region? At this point, I think you need a test that is somewhat aware of the details of Chrome's ICE agent.

lib/twilio/rtc/peerconnection.js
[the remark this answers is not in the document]
ceaglest: Okay. Keep in mind that since the SDK is always offerrer during an ice-restart it should just continue to be the controlling agent (even if previously there were resolved role conflcits) with the same nomination strategy as previously.

tests/unit/device.ts
charliesantos: Type :)

charliesantos:
Thanks @ceaglest . Great feedback. I believe there's a difference in expectation. My understanding during sprint planning, which Anna also has mentioned, is to write tests that will demonstrate the delay, and also tests that shows the delay is fixed after applying the workaround. This is the main motivation factor for these e2e tests. So my thought process was, find a configuration where we can consistently reproduce the significant delay, and verify the delay is greatly reduced after applying the fix. This is also the background for setting CONNECTION_DELAY_THRESHOLD of 1 second. This current setup has been consistent and stable - running on Travis CI with one network interface, using sg1.

About the test scenarios you mentioned - using N network interfaces, testing each delay bucket, packet lost scenario, etc. I agree with you and I think these are important. But I don't think they are important to do now for VoiceJS, considering that we only have one setting that we can change (for this PR) regardless of the varying delays. Also, we are only going to publish this as an experimental feature and possibly for Flex only.

I'm happy to discuss this further and work on them in the future. I just don't think we can do it now, unfortunately. We have other higher priority tasks.

I will respond to the other unrelated comments separately.

tests/integration/icenomination.ts
charliesantos: I think we care about testing the delay from the customer's point of view, which is from gathering to pc-connected. I remember in our slack channel I was doing from checking to connected, but I received a suggestion to measure from gathering instead.

tests/integration/icenomination.ts
charliesantos: We cannot run these test on firefox because:
1) Firefox doesn't support the events that we need.
2) This is disabled on firefox. This type of test covered in our unit tests.

charliesantos:
@ceaglest I updated base on our last discussion. Please check last 3 commits to see incremental changes.

tests/integration/icenomination.ts
ceaglest: Thanks for mentioning the lack of events on FF. 

We should consider adding measurements of Firefox in order to prove that it is not impacted (separate ticket). However, this would require a special measurement technique. It could be using a combination of existing peer-connection events and calls to getStats().

tests/integration/icenomination.ts
ceaglest: 👍 Now matches live discussion. I think we have the correct events to detect the problem under test. Measuring the overall time to media from the customer perspective is different from what we are looking for here.

tests/integration/icenomination.ts
ceaglest: Fixed now that the low delay region is used. 👍

tests/integration/icenomination.ts
ceaglest: This is reasonable given the region finding logic.

charliesantos:
Thanks @ceaglest . I created the following tickets:

CLIENT-7063
Add ability to test DTLS handshake delay with N-network interfaces

CLIENT-7064
Add ability to control delay/RTT

CLIENT-7066
Investigate how to measure dtls handshake duration for browsers without pc connection state support

CLIENT-7065
Add Safari support for integration testing

PR #114, opened 2020-01-24
CLIENT-6880 | Add register/unregister events and open event

ryan-rowland:
This PR:
- Adds register insight event when we receive "ready" VSP message (register success response)
- Adds unregister insight event when we receive "offline" VSP message (unregister success response)
- Adds a new Device event, `open`, indicating the signaling connection is open and ready to place outgoing calls
- Updates the description of the `ready` and `offline` events to reflect that they do not indicate overall Device connectivity, but in fact represent that registration or unregistration for incoming audio calls was successful.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+1.9.8 (In Progress)
+===================
+
+* We now report registration and unregistration for incoming calls to Insights.
+* Added a new Device event, `open`, indicating that the signaling has opened successfully
+  and the Device is ready to place outgoing calls.
+* Updated the `ready` and `offline` event descriptions to reflect that they indicate when 
+  the device has successfully registered or unregistered to listen for incoming calls.
+
 1.9.7 (Dec 6, 2019)
 ===================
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+1.9.8 (In Progress)
```
charliesantos: This is a new API. Should we bump the minor version?

lib/twilio/device.ts
```diff
@@ -1010,6 +1010,7 @@ class Device extends EventEmitter {
    */
   private _onSignalingConnected = (payload: Record<string, any>) => {
     this._region = getRegionShortcode(payload.region) || payload.region;
+    this.emit('open', this);
```
charliesantos: Should we add unit tests for this?

lib/twilio/device.ts
```diff
@@ -1095,6 +1097,7 @@ class Device extends EventEmitter {
   private _onSignalingReady = () => {
     this._log.info('Stream is ready');
     this._status = Device.Status.Ready;
+    this._publisher.info('registration', 'registration', { }, this._activeConnection);
```
charliesantos: This doesn't seem to work because our event publisher requires a connection. We don't have an activeConnection in this case. Same for `unregistration` above.

charliesantos:
Per our discussion at planning, we'll punt this for now.

PR #121, opened 2020-02-12
Changelog for 1.10

charliesantos:
Changelog for 1.10


CHANGELOG.md
```diff
@@ -1,3 +1,25 @@
...
+
+```
+import { getLogger } from 'loglevel';
+const logger = getLogger(Device.packageName);
+
+// Set log level on subsequent page loads and refreshes
+logger.setLevel('DEBUG');
+```
+
+* sdk.twilio.com is now being used for serving the sound files. (CLIENT-7221)
```
aymenn: I think add https://
Does this resolve to a new IP address?

charliesantos: No, this points to a cloudfront distribution.
I'll add https

CHANGELOG.md
```diff
@@ -1,3 +1,25 @@
...
+
+```
+import { getLogger } from 'loglevel';
+const logger = getLogger(Device.packageName);
+
+// Set log level on subsequent page loads and refreshes
+logger.setLevel('DEBUG');
+```
+
+* https://sdk.twilio.com is now being used for serving the sound files. (CLIENT-7221)
```
aymenn: Now I think about it more, we should make a general statement about changing CDN.
What's the plan on updating the instructions for including twilio.js in the changelog?

charliesantos: I was thinking after we release this, we'll update [public changelog](https://www.twilio.com/docs/voice/client/javascript/changelog) with the "Attention" at the top mentioning about the new CDN.

charliesantos:
Was talking with Aymen about 1.10 changelog. We'll copy this over to wagtail and update the debug/warnings section mentioning about loglevel.

PR #141, opened 2020-04-09
[CLIENT-7519] Support for edge in Device.Options

mhuynh5757:
**Contributing to Twilio**

### JIRA link(s):

- [CLIENT-7519](https://issues.corp.twilio.com/browse/CLIENT-7519)

### Description

This PR is a work in progress and should not be merged to `master` yet. It exists for discussion and iteration.

#### TODO
- [x] Check the deprecation warnings and link to docs regarding edge and region changes
- [x] Check the changelog message and link to docs
- [ ] Ensure the generated TSDoc is up-to-date

Add `edge` parameter to `Twilio.Device.Options`, map new edge strings to existing region strings, and add deprecation warnings if `Twilio.Device.Options.region` is used.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,10 +5,12 @@ New Features
 ---------
 
 * Deprecated the `Twilio.Device.Options.region` option for setting up a `Twilio.Device` as part of Phase 1 Regional Twilio.
-Instead of using `Twilio.Device.Options.region`, please use `Twilio.Device.Options.edge`. In the next breaking release, `Twilio.Device.Options.region` will no longer work as intended.
-Please see documentation TODO mhuynh.
+  Instead of using `Twilio.Device.Options.region`, please use `Twilio.Device.Options.edge`.
+  In the next breaking release, `Twilio.Device.Options.region` will no longer work as intended.
 
-The edge that the client connected to can be read from `Twilio.Device` using the getter-function `Twilio.Device.edge`.
+  Please see documentation TODO mhuynh.
+
+  The edge that the client connected to can be read from `Twilio.Device` using the getter-function `Twilio.Device.edge`.
 
 * Added `appName` and `appVersion` fields to Device.options. Pass these strings on Device setup, and they will be passed to Insights. This can
   be used to help debug which of your applications and/or versions an issue began occurring in.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,35 @@ New Features
 
 * Deprecated the `Twilio.Device.Options.region` option for setting up a `Twilio.Device` as part of Phase 1 Regional Twilio.
 Instead of using `Twilio.Device.Options.region`, please use `Twilio.Device.Options.edge`. In a future breaking release, `Twilio.Device.Options.region` will be for a different option.
-Please see TODO mhuynh.
+
+Possible edges are detailed in `lib/regions.ts:Edge`.
+
+```ts
+enum Edge {
+  /**
+   * Public edges
+   */
+  Sydney = 'sydney',
+  SaoPaolo = 'sao-paolo',
+  Dublin = 'dublin',
+  Frankfurt = 'frankfurt',
+  Tokyo = 'tokyo',
+  Singapore = 'singapore',
+  Ashburn = 'ashburn',
+  Umatilla = 'umatilla',
+  Roaming = 'roaming',
+  /**
+   * Interconnect edges
+   */
+  AshburnIx = 'ashburn-ix',
+  SanJoseIx = 'san-jose-ix',
+  LondonIx = 'london-ix',
+  FrankfurtIx = 'frankfurt-ix',
+  SingaporeIx = 'singapore-ix',
+}
+```
+
+Please see documentation TODO mhuynh.
 
 Bug Fixes
 ---------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,15 +4,19 @@
 New Features
 ---------
 
-* Deprecated the `Twilio.Device.Options.region` option for setting up a `Twilio.Device` as part of Phase 1 Regional Twilio.
-  Instead of using `Twilio.Device.Options.region`, please use `Twilio.Device.Options.edge`.
-  In the next breaking release, `Twilio.Device.Options.region` will no longer work as intended.
+* ### Twilio Regional
+  This release includes support for the first phase of the new **Twilio Regional**.
 
-  Please see documentation on [edges](https://www.twilio.com/docs/voice/voip-sdk/edges).
+  **Twilio Regional** brings the possibility to specify a home region where all data will reside and edge connectivity which the Voice SDKs will use to connect to the home region.
 
-  The edge that the client connected to can be read from `Twilio.Device` using the getter-function `Twilio.Device.edge`.
+  This first phase includes edge connectivity and the new parameter `Twilio.Device.Options.edge`.
+  This new parameter supersedes the now deprecated `Twilio.Device.Options.region`.
+  See `Twilio.Device.Options.edge` for migration instructions.
+  The edge that the client connected to can be read from `Twilio.Device` using the read-only property `Twilio.Device.edge`.
 
-* Added `appName` and `appVersion` fields to Device.options. Pass these strings on Device setup, and they will be passed to Insights. This can
+  Please see documentation on [edges](https://www.twilio.com/docs/voice/client/edges).
+
+* Added `appName` and `appVersion` fields to Device.options. Pass these strings on Device setup, and they will be passed to [Insights](https://www.twilio.com/console/voice/insights). This can
   be used to help debug which of your applications and/or versions an issue began occurring in.
   #### Example
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ New Features
   Instead of using `Twilio.Device.Options.region`, please use `Twilio.Device.Options.edge`.
   In the next breaking release, `Twilio.Device.Options.region` will no longer work as intended.
 
-  Please see documentation TODO mhuynh.
+  Please see documentation on [edges](https://www.twilio.com/docs/voice/voip-sdk/edges).
 
   The edge that the client connected to can be read from `Twilio.Device` using the getter-function `Twilio.Device.edge`.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ New Features
 * ### Twilio Regional
   This release includes support for the first phase of the new **Twilio Regional**.
 
-  **Twilio Regional** brings the possibility to specify a home region where all data will reside and edge connectivity which the Voice SDKs will use to connect to the home region.
+  **Twilio Regional** allows developers to specify a region where data is processed and stored and specify an edge indicating where the SDK connects into Twilio.
 
   This first phase includes edge connectivity and the new parameter `Twilio.Device.Options.edge`.
   This new parameter supersedes the now deprecated `Twilio.Device.Options.region`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 1.11.0 (In Progress)
 ===================
 
+New Features
+---------
+
+* Deprecated the `Twilio.Device.Options.region` option for setting up a `Twilio.Device` as part of Phase 1 Regional Twilio.
+Instead of using `Twilio.Device.Options.region`, please use `Twilio.Device.Options.edge`. In a future breaking release, `Twilio.Device.Options.region` will be for a different option.
+Please see TODO mhuynh.
+
 Bug Fixes
 ---------
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ New Features
 ---------
 
 * Deprecated the `Twilio.Device.Options.region` option for setting up a `Twilio.Device` as part of Phase 1 Regional Twilio.
-Instead of using `Twilio.Device.Options.region`, please use `Twilio.Device.Options.edge`. In a future breaking release, `Twilio.Device.Options.region` will be for a different option.
+Instead of using `Twilio.Device.Options.region`, please use `Twilio.Device.Options.edge`. In the next breaking release, `Twilio.Device.Options.region` will no longer work as intended.
 
 Possible edges are detailed in `lib/regions.ts:Edge`.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,34 +6,6 @@ New Features
 
 * Deprecated the `Twilio.Device.Options.region` option for setting up a `Twilio.Device` as part of Phase 1 Regional Twilio.
 Instead of using `Twilio.Device.Options.region`, please use `Twilio.Device.Options.edge`. In the next breaking release, `Twilio.Device.Options.region` will no longer work as intended.
-
-Possible edges are detailed in `lib/regions.ts:Edge`.
-
-```ts
-enum Edge {
-  /**
-   * Public edges
-   */
-  Sydney = 'sydney',
-  SaoPaolo = 'sao-paolo',
-  Dublin = 'dublin',
-  Frankfurt = 'frankfurt',
-  Tokyo = 'tokyo',
-  Singapore = 'singapore',
-  Ashburn = 'ashburn',
-  Umatilla = 'umatilla',
-  Roaming = 'roaming',
-  /**
-   * Interconnect edges
-   */
-  AshburnIx = 'ashburn-ix',
-  SanJoseIx = 'san-jose-ix',
-  LondonIx = 'london-ix',
-  FrankfurtIx = 'frankfurt-ix',
-  SingaporeIx = 'singapore-ix',
-}
-```
-
 Please see documentation TODO mhuynh.
 
 Bug Fixes
--- a/lib/twilio/device.ts
+++ b/lib/twilio/device.ts
@@ -19,11 +19,10 @@ import {
 import Log from './log';
 import { PStream } from './pstream';
 import {
-  defaultEdge,
-  Edge,
-  edgeToRegion,
   getChunderURI,
   getRegionShortcode,
+  Region,
+  regionToEdge,
 } from './regions';
 import { Exception, queryToJson } from './util';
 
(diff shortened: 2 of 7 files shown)


lib/twilio/regions.ts
```diff
@@ -20,13 +19,44 @@ export enum DeprecatedRegion {
   UsVa = 'us-va',
 }
 
+/**
+ * Valid edges.
+ */
+export enum Edge {
```
idelgado, line 25: Might want to add....

```
/**
 * Public Edges
 */
```

lib/twilio/regions.ts
```diff
@@ -39,6 +69,8 @@ export enum Region {
   Ie1Tnx = 'ie1-tnx',
   Jp1 = 'jp1',
   Sg1 = 'sg1',
+  Sg1Ix = 'sg1-ix',
+  Sg1Tnx = 'sg1-tnx',
```
charliesantos, line 73: Are these new? I thought we already have sg interconnect before?

lib/twilio/regions.ts
```diff
@@ -136,6 +243,86 @@ export function getRegionURI(region?: string, onDeprecated?: (newRegion: string)
...
+  if (onDeprecated && deprecatedMessages.length) {
+    setTimeout(() => onDeprecated(deprecatedMessages.join(' ')));
+  }
+
+  if (chunderRegion === defaultRegion) {
+    return defaultChunderUri;
+  }
+
+  return `chunderw-vpc-gll-${chunderRegion}.twilio.com`;
+}
```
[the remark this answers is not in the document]
charliesantos: Actually, what do you think about this? I think this covers the cases we're trying to do. Can you please check if this make sense, and if so, are you ok using it? It should shrink down this method.
```ts
  const deprecatedMessages: string[] = [];
  if (region) {
    deprecatedMessages.push('Regions are deprecated in favor of edges. Please see TODO.');
  }
  // Let's do the mapping first, from edge to region
  let chunderRegion = region || (Object.values(Edge) as string[]).includes(edge) ? edgeToRegion[edge as Edge] : edge;

  // Check if region is deprecated
  const newRegion = deprecatedRegions[region as DeprecatedRegion];
  if (newRegion) {
    deprecatedMessages.push(`Region ${chunderRegion} is deprecated, please use ${newRegion}.`);
    chunderRegion = newRegion;
  }

  // Return and trigger callback
  if (onDeprecated && deprecatedMessages.length) {
    setTimeout(() => onDeprecated(deprecatedMessages.join('\n')));
  }
  if (chunderRegion === defaultRegion) {
    return defaultChunderUri;
  }
  return `chunderw-vpc-gll-${chunderRegion}.twilio.com`;
```

charliesantos:
@idelgado , do we need to change any insights schema? Or just keep it as is? We are currently sending `selected_region` and `region` values.

@ryan-rowland can you please also take a look at this PR?

lib/twilio/device.ts
```diff
@@ -580,6 +580,12 @@ class Device extends EventEmitter {
       throw new InvalidArgumentError('Token is required for Device.setup()');
     }
 
+    const regionURI = getChunderURI(
```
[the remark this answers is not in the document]
mhuynh5757: I was thinking of moving it here because we throw an error in `getChunderURI` and I didn't want to initialize anything in that case. From a discussion with @idelgado , we were thinking of just failing when both `edge` and `region` are passed to prevent ambiguity, instead of giving edge priority over region.

lib/twilio/regions.ts
mhuynh5757: When we added `sg interconnect` support, we had moved to using the string template instead of the mapping, so the mapping was deprecated and not updated. Since we have to use the mapping again for edges, I just added `sg-ix` and `tnx` to the mapping to update it.

lib/twilio/device.ts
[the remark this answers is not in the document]
charliesantos: Ok got it.

mhuynh5757:
After some discussion, we're thinking of skipping having a private `_edge` class member because `edge` does not have the same implications as `region` does. `edge` is only an option (the desired entry point into the Twilio network), and will never change, unlike `region`. For example, passing a `region` of `gll` to `Twilio.Device` may result in the signaling stack connecting to `us1`, so we reflect that in the private class member.

idelgado:
@charliesantos yes. lets consider what changes are required for insights. I think we should propose a new value `edge` that we publish to insights when the SDK developer opts to use the new option in the SDK or when the developer does not provide any options and uses the default "roaming".

lib/twilio/regions.ts
```diff
@@ -136,6 +246,79 @@ export function getRegionURI(region?: string, onDeprecated?: (newRegion: string)
...
+ *
+ * @private
+ * @param edge - The edge.
+ * @param region - The region shortcode.
+ * @param [onDeprecated] - A callback containing the deprecation message to be
+ *   warned when the passed parameters are deprecated.
+ */
+export function getChunderURI(
+  edge: any,
+  region: any,
```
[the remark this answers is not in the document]
mhuynh5757: I was thinking that we could use `any` here because we won't know what type the customer will pass for the option. Using the `typeof` forces `edge` and `region` such that they have strictly `string | undefined` types after the type-guard for the rest of the function.

I'm on the fence about this one though, because it implies the function is _ok_ with passing `any` as a parameter.

Which would you prefer?

lib/twilio/regions.ts
```diff
@@ -136,6 +246,79 @@ export function getRegionURI(region?: string, onDeprecated?: (newRegion: string)
...
+ * As of CLIENT-7519, Regions are deprecated in favor of edges as part of
+ * Phase 1 Regional.
+ *
+ * @private
+ * @param edge - The edge.
+ * @param region - The region shortcode.
+ * @param [onDeprecated] - A callback containing the deprecation message to be
+ *   warned when the passed parameters are deprecated.
+ */
+export function getChunderURI(
```
charliesantos, line 261: Since we're using this now, why did we not remove getRegionURI above?

charliesantos:
Implementation is good. There's more work on the documentation part. Who is taking care of the documentation? Should we merge this and take care of documentation in a separate PR? We have 2 tickets that are blocked by this PR.

lib/twilio/regions.ts
```diff
@@ -104,36 +142,181 @@ const regionURIs: Record<Region, string> = {
...
+    ? defaultChunderRegionURI
+    : `chunderw-vpc-gll-${region}.twilio.com`;
+}
+
+/**
+ * The default chunder URI to connect to, should map to edge `roaming`.
+ * @constant
+ * @private
+ */
+export const defaultChunderEdgeURI: string = 'voice.roaming.twilio.com';
```
charliesantos, line 239: `voice-js.roaming.twilio.com` per https://docs.google.com/document/d/1efmXWxlUOWUWl5BNmrUiAZVpI2lTojc-vw66fMkb87w/edit?ts=5e4d58a5#heading=h.hry0rjrq6ra7

lib/twilio/regions.ts
```diff
@@ -104,36 +142,181 @@ const regionURIs: Record<Region, string> = {
...
+
+/**
+ * String template for an edge chunder URI
+ * @param edge - The edge.
+ */
+function createChunderEdgeUri(edge: string): string {
+  return `voice.${edge}.twilio.com`;
```
charliesantos, line 246: `voice-js.${edge}.twilio.com` per https://docs.google.com/document/d/1efmXWxlUOWUWl5BNmrUiAZVpI2lTojc-vw66fMkb87w/edit?ts=5e4d58a5#heading=h.hry0rjrq6ra7

CHANGELOG.md
```diff
@@ -1,8 +1,16 @@
 1.11.0 (In Progress)
 ===================
 
-Features
---------
+New Features
+---------
+
```
aymenn, line 6: This release includes support for the first phase of the new Twilio Regional. Twilio Regional brings the possibility to specify a home region where all data will reside and edge connectivity which the Voice SDKs will use to connect to the home region. This first phase includes edge connectivity and the new param `Twilio.Device.Options.edge`. This new parameter superseded the now deprecated `Twilio.Device.Options.region`. See `Twilio.Device.Options.edge` for migration instructions.

@ryan-rowland @idelgado would love your input on this changelog

lib/twilio/device.ts
```diff
@@ -532,6 +549,15 @@ class Device extends EventEmitter {
    * if not connected.
    */
   region(): string {
+    this._log.warn(
+      'The value that `Twilio.Device.region` returns is deprecated. ' +
```
aymenn, line 553: Replace paragraph with: region is deprecated and will be removed in the next major release. Please use edge instead

CHANGELOG.md
idelgado: Read and made a minor edit.

CHANGELOG.md
```diff
@@ -1,8 +1,16 @@
...
+* Deprecated the `Twilio.Device.Options.region` option for setting up a `Twilio.Device` as part of Phase 1 Regional Twilio.
+  Instead of using `Twilio.Device.Options.region`, please use `Twilio.Device.Options.edge`.
+  In the next breaking release, `Twilio.Device.Options.region` will no longer work as intended.
```
idelgado, line 9: > In the next major API release, `Twilio.Device.Options.region` will no longer be available.

CHANGELOG.md
```diff
@@ -1,8 +1,16 @@
...
+
+  Please see documentation on [edges](https://www.twilio.com/docs/voice/voip-sdk/edges).
+
+  The edge that the client connected to can be read from `Twilio.Device` using the getter-function `Twilio.Device.edge`.
```
idelgado, line 13: > can be retrieved from ... using the getter method ...

CHANGELOG.md
```diff
@@ -1,8 +1,16 @@
...
 
 * Added `appName` and `appVersion` fields to Device.options. Pass these strings on Device setup, and they will be passed to Insights. This can
```
idelgado, line 15: > Developers can provide their application name and version to help identify issue that may be specific to a specific application version.

lib/twilio/device.ts
```diff
@@ -272,6 +273,11 @@ class Device extends EventEmitter {
    */
   private _connectionSinkIds: string[] = ['default'];
 
+  /**
+   * The edge the {@link Device} is connected to.
```
idelgado, line 277: > The name of the edge that the {@link Device} is connected to.

lib/twilio/device.ts
```diff
@@ -491,6 +496,18 @@ class Device extends EventEmitter {
...
 
+  /**
+   * Get the {@link Edge} string the {@link Device} is currently connected to,
+   * or 'offline' if not connected.
+   */
+  get edge(): string | null {
+    return this.isInitialized
+      ? typeof this._edge === 'string'
+        ? this._edge
+        : 'offline'
```
idelgado, line 507: Why `offline` vs `null` or `undefined`?

lib/twilio/device.ts
```diff
@@ -491,6 +496,18 @@ class Device extends EventEmitter {
     this._disconnectAll();
   }
 
+  /**
+   * Get the {@link Edge} string the {@link Device} is currently connected to,
```
idelgado, line 500: > Returns the {@link Edge} value the {@link Device} is currently connected to. The value will be 'offline' when the Device is offline.

lib/twilio/device.ts
```diff
@@ -1471,6 +1505,15 @@ namespace Device {
      */
     dscp?: boolean;
 
+    /**
+     * The edge value corresponds to the geographic location that the client
+     * will use to connect to Twilio infrastructure. The default value is
+     * "roaming" which automatically selects an edge based on the latency of the
+     * client relative to available edges. You may not specify both `edge` and
```
idelgado, line 1512: > Specifying both `edge` and `region` in the Device options will result in an `IllegalStateException`.

lib/twilio/regions.ts
```diff
@@ -104,36 +142,174 @@ const regionURIs: Record<Region, string> = {
...
+  edge: string | undefined,
+  region: string | undefined,
+  onDeprecated?: (message: string) => void,
+): string {
+  if (
+    (typeof region !== 'string' && typeof region !== 'undefined') ||
+    (typeof edge !== 'string' && typeof edge !== 'undefined')
+  ) {
+    throw new InvalidArgumentError(
+      'If `region` or `edge` are defined, they must of type `string`.',
```
idelgado, line 264: I would throw separate messags for this.

`throw new InvalidArgumentError("region must be defined as a string")` 

`throw new InvalidArgumentError("edge must be defined as a string")`

lib/twilio/regions.ts
```diff
@@ -104,36 +142,174 @@ const regionURIs: Record<Region, string> = {
...
-    if (onDeprecated) {
-      // Don't let this callback affect script execution.
-      setTimeout(() => onDeprecated(newRegion));
+  const deprecatedMessages: string[] = [];
+
+  let uri: string = defaultChunderRegionURI;
+
+  if (region && edge) {
+    throw new InvalidArgumentError(
+      'Defining both a `region` and an `edge` in `Twilio.Device.Options` is ' +
```
idelgado, line 274: > You cannot specify `region` when `edge` is specified in `Twilio.Device.Options`.

lib/twilio/device.ts
ryan-rowland: I think this is an artifact of `region` which had an `offline` region. I'm +1 to move to `null`.

CHANGELOG.md
ryan-rowland: From the user's perspective this is neither a function nor a method. It's just a property. It's accessed eg console.log(Twilio.Device.edge) rather than console.log(Twilio.Device.edge()). I'd call it a property or a read-only property if you want to be specific.

CHANGELOG.md
ryan-rowland: This was actually from my PR but we can update the verbiage here. 👍

ryan-rowland: @idelgado To be clear you're only talking about replacing "This can be used...." and not the first line which describes the field names, correct?

lib/twilio/device.ts
ryan-rowland: Interestingly, this is described as "string | null" but never returns null. +1 to dropping the 'offline' string, and let's make sure we update this docblock to reflect that it will be null if not connected.

lib/twilio/device.ts
```diff
@@ -853,11 +884,11 @@ class Device extends EventEmitter {
...
-    const stream: IPStream = this.stream;
-    if (stream) {
-      setIfDefined('gateway', stream.gateway);
-      setIfDefined('region', stream.region);
-    }
+    setIfDefined('gateway', this.stream && this.stream.gateway);
+    setIfDefined('selected_region', this.options.region);
+    setIfDefined('region', this.stream && this.stream.region);
+    setIfDefined('selected_edge', this.options.edge);
+    setIfDefined('edge', this._edge || this._region);
```
ryan-rowland, line 891: Why do we use `this.stream.region` for region but `this._region` for the edge fallback? What's the difference between the two?

lib/twilio/device.ts
ryan-rowland: Shouldn't this be `InvalidArgumentException`?

lib/twilio/regions.ts
```diff
@@ -104,36 +142,174 @@ const regionURIs: Record<Region, string> = {
...
+      'unsupported.',
+    );
+  } else if (region) {
+    let chunderRegion = region;
+
+    deprecatedMessages.push(
+      'Regions are deprecated in favor of edges. Please see this page for ' +
+      'documentation: https://www.twilio.com/docs/voice/voip-sdk/edges.',
```
ryan-rowland, line 282: Update URL here as well based on @aymenn's feedback above

lib/twilio/device.ts
mhuynh5757: Talking with @charliesantos a bit about this, we decided to leave `this.stream.region` for the `region` insights field to maintain compatibility because we have always been sending this value to insights. `this.stream.region` is the AWS region we get from signaling (i.e. something like `ASIAPAC_SINGAPORE`) where `this._region` is the shortcode version of that (i.e. `sg1`).

CHANGELOG.md
```diff
@@ -1,10 +1,22 @@
...
+
+* ### Twilio Regional
+  This release includes support for the first phase of the new **Twilio Regional**.
+
+  **Twilio Regional** brings the possibility to specify a home region where all data will reside and edge connectivity which the Voice SDKs will use to connect to the home region.
+
+  This first phase includes edge connectivity and the new parameter `Twilio.Device.Options.edge`.
+  This new parameter supersedes the now deprecated `Twilio.Device.Options.region`.
+  See `Twilio.Device.Options.edge` for migration instructions.
+  The edge that the client connected to can be read from `Twilio.Device` using the read-only property `Twilio.Device.edge`.
```
idelgado, line 15: This looks like a method rather than a property from an API perspective.

CHANGELOG.md
```diff
@@ -1,10 +1,22 @@
...
 
-Features
---------
+New Features
+---------
+
+* ### Twilio Regional
+  This release includes support for the first phase of the new **Twilio Regional**.
+
+  **Twilio Regional** brings the possibility to specify a home region where all data will reside and edge connectivity which the Voice SDKs will use to connect to the home region.
```
idelgado, line 10: I actually wonder if we should avoid using the term home. Wdyt @aymenn ?

How about...

```
**Twilio Regional** allows developers to specify a region where data is processed and stored and specify an edge indicating where the SDK connects into Twilio. 
```

lib/twilio/device.ts
```diff
@@ -491,6 +496,14 @@ class Device extends EventEmitter {
     this._disconnectAll();
   }
 
+  /**
+   * Returns the {@link Edge} value the {@link Device} is currently connected
+   * to. The value will be `null` when the {@link Device} is offline.
+   */
+  get edge(): string | null {
```
idelgado, line 503: @ryan-rowland null or undefined? Can you remind me when it's advisable to use null vs undefined?

CHANGELOG.md
charliesantos: Implementation wise, it is a getter method. But from an SDK user perspective, it is a property of Device.

lib/twilio/device.ts
ryan-rowland: I think I've gone back and forth with @manjeshbhargav on this, but I'd say when we are intentionally saying there is an absence of a value it should be `null`. Most of our API should then be `null` if not set, unless that property is a result of a user / external input, and that input is not defined.

charliesantos:
Closing this and will be submitted to 1.11 branch, per team's discussion this morning.

PR #150, opened 2020-04-23
CLIENT-7583 STIR/SHAKEN

ryan-rowland:
This adds a CallerId field onto Connection based on [this API spec](https://docs.google.com/document/d/1Wx6n4zQ-_IpEl1VtQCuiyIusqVIG4EU3tpHw-9rL1Mk/edit).

Also see the corresponding [VSP update PR](https://code.hq.twilio.com/client/voice-signaling-protocol/pull/6).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,31 @@
+1.12.0 (In progress)
+====================
+
+New Features
+---------
+
+### CallerInfo
+
+* Added a `Connection.callerInfo` field that will hold useful information about the caller when the
+  caller is a PSTN number. Currently, one new field is available:
+    1. `CallerInfo.isVerified` -- A boolean indicating whether or not Twilio was able to verify
+        whether the caller is authorized to use the number that this call is from. If true, it
+        is safe to trust that the caller is who they claim to be. If this is false, it does not
+        mean that the call is fraudulent, only that Twilio was not able to verify authenticity.
+        Most legitimate calls at the time of implementation will not be verifiable as most
+        numbers are not set up to utilize the underlying STIR/SHAKEN protocol.
+
+#### Example
+```ts
+device.on('incoming', connection => {
+  if (connection.callerInfo && connection.callerInfo.isVerified) {
+    showVerifiedBadge();
+  }
+});
+```
+
 1.11.0 (May 21, 2020)
-===================
+=====================
 
 New Features
 ---------
--- a/lib/twilio/connection.ts
+++ b/lib/twilio/connection.ts
@@ -95,6 +95,12 @@ class Connection extends EventEmitter {
   static toString = () => '[Twilio.Connection class]';
 
   /**
+   * Describes the phone number that is initiating the call.
+   * Null if this call is outbound from this Client, or if the caller is not from PSTN.
+   */
+  readonly callerInfo: Connection.CallerInfo | null;
+
+  /**
    * The custom parameters sent to (outgoing) or received by (incoming) the TwiML app.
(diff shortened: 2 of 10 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,31 @@
...
+    1. `CallerInfo.isVerified` -- A boolean indicating whether or not Twilio was able to verify
+        whether the caller is authorized to use the number that this call is from. If true, it
+        is safe to trust that the caller is who they claim to be. If this is false, it does not
+        mean that the call is fraudulent, only that Twilio was not able to verify authenticity.
+        Most legitimate calls at the time of implementation will not be verifiable as most
+        numbers are not set up to utilize the underlying STIR/SHAKEN protocol.
+
+#### Example
+```ts
+device.on('invite', connection => {
```
charliesantos, line 20: This is a new event correct? In addition to adding this example, should we mention in the changelog? Something like
```
* Added a new `Device.on('invite')` event which is raised when blah blah
```

lib/twilio/connection.ts
```diff
@@ -261,6 +267,15 @@ class Connection extends EventEmitter {
 
     this._direction = this.parameters.CallSid ? Connection.CallDirection.Incoming : Connection.CallDirection.Outgoing;
 
+    if (this._direction === Connection.CallDirection.Incoming && this.parameters) {
+      const isFromPSTN: boolean = /^\+?[\d-\(\) ]+$/.test(this.parameters.From);
+      this.callerInfo = isFromPSTN
+        ? { isVerified: this.parameters.StirStatus === 'TN-Validation-Passed-A' }
```
charliesantos, line 273: Not sure if related, but curious to know how this works if the caller is coming from a conference? Will chunder know it is coming from a PSTN via a conference? Or we don't care about this scenario for now?

CHANGELOG.md
ryan-rowland: haha No, I just have been spending too much time with RTC SDK. I'll fix to `incoming`.

lib/twilio/connection.ts
ryan-rowland: @charliesantos The current API is to only show calls coming in from PSTN, but this is a good question, since most of our call center customers use conferences over calls. I'll ask for clarification.

tests/unit/connection.ts
```diff
@@ -102,6 +102,141 @@ describe('Connection', function() {
...
+        let callerInfo: Connection.CallerInfo;
+        if (conn.callerInfo !== null) {
+          callerInfo = conn.callerInfo;
+          assert.equal(callerInfo.isVerified, false);
+        } else {
+          throw Error('callerInfo object null, but expected to be populated');
+        }
+      });
+
+      it('should populate the .callerInfo fields appropriately when StirStatus is no-valifadation', () => {
```
mhuynh5757, line 196: sp: validation

CHANGELOG.md
```diff
@@ -1,5 +1,31 @@
...
+====================
+
+New Features
+---------
+
+### CallerInfo
+
+* Added a `Connection.callerInfo` field that will hold useful information about the caller when the
+  caller is a PSTN number. Currently, one new field is available:
+    1. `CallerInfo.isVerified` -- A boolean indicating whether or not Twilio was able to verify
```
idelgado, line 11: Typically the IDL language is used in the documentation and API. Otherwise there's little value in doing the IDL in the first place. We should reconcile the two. Personally, I prefer what we have in the IDL.

PR #154, opened 2020-04-27
CLIENT-7562 | Adding edge fallback support

charliesantos:
**Contributing to Twilio**

### JIRA link(s):

- [CLIENT-7562](https://issues.corp.twilio.com/browse/CLIENT-7562)

### Description

Implemented regional edge fallback. See [CLIENT-7562](https://issues.corp.twilio.com/browse/CLIENT-7562) ticket for specs.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,22 @@ New Features
   See `Twilio.Device.Options.edge` for migration instructions.
   The edge that the client connected to can be read from `Twilio.Device` using the read-only property `Twilio.Device.edge`.
 
-  Please see documentation on [edges](https://www.twilio.com/docs/voice/client/edges).
+  Please see documentation on [edges](https://www.twilio.com/docs/voice/client/edges) for more details.
+
+  **Example**  
+  ```ts
+  const device = new Device(token, { edge: 'ashburn' });
+  ```
+
+  **Edge Fallback**  
+  The new parameter `Twilio.Device.Options.edge`, can also be used to provide auto fallback functionality. This is enabled if the `edge` parameter is provided as an array with more than one edge value, sorted by priority order where the first element has the highest priority and the last element has the lowest priority.
+
+  If the `edge` parameter is provided as an array, adding `roaming` as one of the values will cause `device.setup()` to throw an exception. Please see [edges](https://www.twilio.com/docs/voice/client/edges) for other valid edge values.
+
+  **Example**
+  ```ts
+  const device = new Device(token, { edge: ['ashburn', 'sydney', 'dublin'] });
+  ```
 
 * Added `appName` and `appVersion` fields to Device.options. Pass these strings on Device setup, and they will be passed to [Insights](https://www.twilio.com/console/voice/insights). This can
   be used to help debug which of your applications and/or versions an issue began occurring in.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,17 +21,13 @@ New Features
   const device = new Device(token, { edge: 'ashburn' });
   ```
 
-  ### Edge Fallback
+  ### Edge Fallback Support
 
-  The new parameter `Twilio.Device.Options.edge`, can also be used to provide auto fallback functionality. This is enabled if the `edge` parameter is provided as an array with more than one edge value, sorted by priority order where the first element has the highest priority and the last element has the lowest priority.
-
-  If the `edge` parameter is provided as an array, adding `roaming` as one of the values will cause `device.setup()` to throw an exception. Please see [edges](https://www.twilio.com/docs/voice/client/edges) for other valid edge values.
-
-  The fallback behavior is only applicable if edge parameter is supplied as an array. During initialization, the SDK will use the first edge value in the edges array. If the SDK encounters a WebSocket error code of [1006 or 1015](https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent), the SDK will attempt to reconnect using the next value in the edges array with an initial delay of 1+random(5) seconds and will cycle through all the edge values until a connection is established. An exponential backoff will be applied in between connection attempts with a maximum delay of 20 seconds. This maximum delay can be changed using `Twilio.Device.Options.backoffMaxMs`;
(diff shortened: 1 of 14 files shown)


CHANGELOG.md
```diff
@@ -14,7 +14,25 @@ New Features
...
+
+  If the `edge` parameter is provided as an array, adding `roaming` as one of the values will cause `device.setup()` to throw an exception. Please see [edges](https://www.twilio.com/docs/voice/client/edges) for other valid edge values.
+
+  The fallback behavior is only applicable if edge parameter is supplied as an array. During initialization, the SDK will use the first edge value in the edges array. If the SDK encounters a WebSocket error code of [1006 or 1015](https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent), the SDK will attempt to reconnect using the next value in the edges array with an initial delay of 1+random(5) seconds and will cycle through all the edge values until a connection is established. An exponential backoff will be applied in between connection attempts with a maximum delay of 20 seconds. This maximum delay can be changed using `Twilio.Device.Options.backoffMaxMs`;
+
+  **Example**
+  ```ts
+  const device = new Device(token, { edge: ['ashburn', 'sydney', 'dublin'] });
+  ```
```
ryan-rowland, line 35: cc @aymenn Please review changelog entry

CHANGELOG.md
```diff
@@ -14,7 +14,25 @@ New Features
...
 
-  Please see documentation on [edges](https://www.twilio.com/docs/voice/client/edges).
+  Please see documentation on [edges](https://www.twilio.com/docs/voice/client/edges) for more details.
+
+  **Example**  
+  ```ts
+  const device = new Device(token, { edge: 'ashburn' });
+  ```
+
+  ### Edge Fallback
```
aymenn, line 24: ### Edge Fallback Support

Deployments designed to connect to multiple Twilio regions can take advantage of the new edge fallback mechanism. To enable the edge fallback, specify an array of edge names to `Twilio.Device.Options.edge`. When enabled and a connection failure is encountered, the SDK will reattempt the connection to the next region in the list. For more details about how the fallback works, see `Twilio.Device.Options.edge` documentation.

**Example**
  ```ts
  const device = new Device(token, { edge: ['ashburn-ix', 'san-jose-ix' ] });
  ```

CHANGELOG.md
```diff
@@ -14,7 +14,25 @@ New Features
...
+
+  If the `edge` parameter is provided as an array, adding `roaming` as one of the values will cause `device.setup()` to throw an exception. Please see [edges](https://www.twilio.com/docs/voice/client/edges) for other valid edge values.
+
+  The fallback behavior is only applicable if edge parameter is supplied as an array. During initialization, the SDK will use the first edge value in the edges array. If the SDK encounters a WebSocket error code of [1006 or 1015](https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent), the SDK will attempt to reconnect using the next value in the edges array with an initial delay of 1+random(5) seconds and will cycle through all the edge values until a connection is established. An exponential backoff will be applied in between connection attempts with a maximum delay of 20 seconds. This maximum delay can be changed using `Twilio.Device.Options.backoffMaxMs`;
```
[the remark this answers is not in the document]
charliesantos: @aymenn are you referring to this link? https://www.twilio.com/docs/voice/client/edges
I think that is a placeholder page in wagtail and planning to get filled in before we ship regional.
@mhuynh5757 is there a draft in that page or is it still empty?

tests/docker/proxy/server.js
```diff
@@ -186,6 +190,31 @@ class DockerProxyServer {
     });
   }
 
+  async _generateCapabilityToken(shouldInvalidate) {
```
idelgado, line 193: New code shouldn't use capability token. Can we use access token instead?

idelgado:
To validate behavior is it possible to run a localhost server that would result in the 1006 or 1015? I noticed we are primarily mocking them but are we certain these are the codes we would see when an edge fails?

tests/docker/proxy/server.js
charliesantos: This is used by [one of my tests](https://github.com/twilio/twilio-client.js/blob/CLIENT-7562/tests/integration/edgefallback.ts#L70) where using an invalid token generates a server disconnect error that is not fallback-able. The rest of the tests uses access token.

idelgado: Does the access token way not result in a disconnect error that is not fallback-able too?

charliesantos:
@idelgado 
> Is there a way through insights to determine when fallback occurs?

Not with the current implementation since there won't be an active call when fallback occurs. We thought about sending this information but insights is currently call centric and I don't think we can send an event without a callsid. I believe we've discussed with Ryan and Aymen before. We can have that discussion with the team if you think this is possible.

> To validate behavior is it possible to run a localhost server that would result in the 1006 or 1015? I noticed we are primarily mocking them but are we certain these are the codes we would see when an edge fails?

We are doing this already with the e2e tests (using docker). Except for 1015, which I'm not sure how to add in the tests. Our tests covers handling 1006, which follows the same code path for when a 1015 error happens. Do you think it's worth investigating how to generate 1015 even though the code path is the same as 1006 error?

tests/docker/proxy/server.js
charliesantos: Just to clarify, I need a way to generate a server error which closes the connection _**while**_ I'm already connected to the server. This error and closing the connection should be initiated by the server.

The access token way behaves differently on the server. If the access token is invalid, we get an error but it doesn't close the websocket connection.

But with invalid capability token, the server generates an error and also terminates the connection.

idelgado:
@charliesantos @ryan-rowland can we use the temp_call_sid for Insights?

charliesantos:
@idelgado we can update the SDK so that it can send events without a callsid. It's not hard but it's also a fair amount of work to support that. I prefer doing that instead of sending a dummy callsid.
Are you thinking of creating new insight events? One on disconnect, and another one on reconnected. The events will contain the URI or edge location.

aymenn:
@charliesantos confirmed we log to the console when doing a fallback. I think this is sufficient.

charliesantos:
Here's the slack discussion thread https://twilio.slack.com/archives/C0AQK5J5Q/p1588691293383900?thread_ts=1588689561.380000&cid=C0AQK5J5Q

PR #156, opened 2020-04-28
Fix qe test trigger

ruescar:


.circleci/config.yml
```diff
@@ -3,19 +3,19 @@ version: 2.1
 jobs:
   trigger-qe-tests:
     docker:
-    - image: circleci/node:latest
+      - image: circleci/node:latest
```
DenissSazanovs, line 6: If we adding space here, should we do it everywhere in similar situations?

ruescar: I think in this particular case, config would still be valid, but it's just luck. When you use bigger definition for jobs with generic executors etc, you may get an invalid config file and will throw and error. I wouldn't change other places if they are working, but yes generally speaking, indentation can break a circleci config in some cases.

PR #166, opened 2020-05-12
CLIENT-7652 | Don't fallback edge when coming from connected

charliesantos:
**Contributing to Twilio**

### JIRA link(s):

- [CLIENT-7652](https://issues.corp.twilio.com/browse/CLIENT-7652)

### Description

Still needs tests but an initial review on the solution would help. I'm not 100% happy with the approach but this seems to be the simplest solution I found considering different browser behaviors. I intentionally made it verbose for readability.

EDIT: This is now ready. I removed isSafari check and added a fallback counter instead to cover all browsers.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/lib/twilio/wstransport.ts
+++ b/lib/twilio/wstransport.ts
@@ -234,23 +234,6 @@ export default class WSTransport extends EventEmitter {
   }
 
   /**
-   * Check if this connection was previously connected
-   */
-  private _checkWasConnected(): boolean {
-    return (
-      // Only in Safari and certain Firefox versions, on network interruption, websocket drops right away with 1006
-      // Let's check current state if it's open, meaning we should not fallback
-      // because we're coming from a previously connected session
-      this.state === WSTransportState.Open ||
-
-      // But on other browsers, websocket doesn't drop
-      // but our heartbeat catches it, setting the internal state to "Connecting".
-      // With this, we should check the previous state instead.
-      this._previousState === WSTransportState.Open
-    );
-  }
-
-  /**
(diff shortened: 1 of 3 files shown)


ryan-rowland:
Change looks good, we should add a test though.

lib/twilio/wstransport.ts
```diff
@@ -90,6 +90,11 @@ export default class WSTransport extends EventEmitter {
    */
   private _connectTimeoutMs?: number;
 
+  /**
+   * Keeps track of fallbackable errors happened in this connection
+   */
+  private _fallbackableErrorCount: number = 0;
```
ryan-rowland, line 96: How about
```
  /**
   * Whether we should attempt to fallback if we receive an applicable error.
   */
  private _shouldFallback:  boolean = true;
```

lib/twilio/wstransport.ts
```diff
@@ -337,7 +347,27 @@ export default class WSTransport extends EventEmitter {
...
+        this._previousState === WSTransportState.Open) &&
+
+        // We also need to make sure this is the first fallbackable error
+        this._fallbackableErrorCount === 1) {
+          shouldFallback = false;
+      }
+
+      if (shouldFallback) {
+        this._moveUriIndex();
+      }
```
ryan-rowland, line 370: Then this could be something like:

if (this._shouldFallback) {
  this._moveUriIndex();
  this._shouldFallback = false;
}

PR #169, opened 2020-05-19
CLIENT-7627 | Remove mic dependency on preflight

charliesantos:
**Contributing to Twilio**

### JIRA link(s):

- [CLIENT-7627](https://issues.corp.twilio.com/browse/CLIENT-7627)

### Description

See ticket details

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/PREFLIGHT.md
+++ b/PREFLIGHT.md
@@ -20,6 +20,7 @@ The `options` parameter is a JavaScript object containing configuration settings
 | `codecPreferences` | `['pcmu', 'opus']` | An ordered list of preferred codecs. |
 | `debug` | `false` | Can be `true` or `false`. Set this property to true to enable debug logging in your browser console. |
 | `edge` | `roaming` | Specifies which Twilio `edge` to use when initiating the test call. Please see documentation on [edges](https://www.twilio.com/docs/voice/client/edges). |
+| `ignoreMicInput` | `false` | If set to `true`, the test call will ignore microphone input and will use a default audio file. If set to `false`, the test call will capture the audio from the microphone. Setting this to `true` is only supported on Chrome and will throw an `InputStreamNotSupportedError` on other browsers |
 | `signalingTimeoutMs` | `10000` | Ammount of time to wait for setting up signaling connection. |
 
 Events
--- a/PREFLIGHT.md
+++ b/PREFLIGHT.md
@@ -20,7 +20,7 @@ The `options` parameter is a JavaScript object containing configuration settings
 | `codecPreferences` | `['pcmu', 'opus']` | An ordered list of preferred codecs. |
 | `debug` | `false` | Can be `true` or `false`. Set this property to true to enable debug logging in your browser console. |
 | `edge` | `roaming` | Specifies which Twilio `edge` to use when initiating the test call. Please see documentation on [edges](https://www.twilio.com/docs/voice/client/edges). |
-| `ignoreMicInput` | `false` | If set to `true`, the test call will ignore microphone input and will use a default audio file. If set to `false`, the test call will capture the audio from the microphone. Setting this to `true` is only supported on Chrome and will throw an `InputStreamNotSupportedError` on other browsers |
+| `fakeMicInput` | `false` | If set to `true`, the test call will ignore microphone input and will use a default audio file. If set to `false`, the test call will capture the audio from the microphone. |
 | `signalingTimeoutMs` | `10000` | Ammount of time to wait for setting up signaling connection. |
 
 Events
--- a/PREFLIGHT_TWIML.md
+++ b/PREFLIGHT_TWIML.md
@@ -1,7 +1,7 @@
 TwiML App - Record and Play
 ===========================
 
-If `PreflightTest.Options.ignoreMicInput` is set to `false`, `Device.testPreflight(token, options)` API requires a [token](https://www.twilio.com/docs/iam/access-tokens) with a TwiML app that can record an audio from a microphone and the ability to play the recorded audio back to the browser. In order to achieve this, we need two TwiML endpoints: one to capture and record the audio, and another one to play the recorded audio.
+If `PreflightTest.Options.fakeMicInput` is set to `false`, `Device.testPreflight(token, options)` API requires a [token](https://www.twilio.com/docs/iam/access-tokens) with a TwiML app that can record an audio from a microphone and the ability to play the recorded audio back to the browser. In order to achieve this, we need two TwiML endpoints: one to capture and record the audio, and another one to play the recorded audio.
 
 TwiML Bins
 ----------
@@ -49,7 +49,7 @@ On that same page, open the TwiML app that you just created by clicking on it an
 TwiML App - Echo
 ================
 
-If `PreflightTest.Options.ignoreMicInput` is set to `true`, `Device.testPreflight(token, options)` API requires a [token](https://www.twilio.com/docs/iam/access-tokens) with a single TwiML app that can capture and play an audio. Following the [previous steps](#twiml-bins), create a TwiML Bin using the following template, and name it `Echo`.
+If `PreflightTest.Options.fakeMicInput` is set to `true`, `Device.testPreflight(token, options)` API requires a [token](https://www.twilio.com/docs/iam/access-tokens) with a single TwiML app that can capture and play an audio. Following the [previous steps](#twiml-bins), create a TwiML Bin using the following template, and name it `Echo`.
 
 ```xml
 <?xml version="1.0" encoding="UTF-8"?>
--- a/PREFLIGHT_TWIML.md
+++ b/PREFLIGHT_TWIML.md
@@ -1,7 +1,7 @@
-Preflight Test TwiML App
-=========================
+TwiML App - Record and Play
+===========================
 
-`Device.testPreflight(token, options)` API requires a [token](https://www.twilio.com/docs/iam/access-tokens) with a TwiML app that can record an audio from a microphone and the ability to play the recorded audio back to the browser. In order to achieve this, we need two TwiML endpoints: one to capture and record the audio, and another one to play the recorded audio.
+If `PreflightTest.Options.ignoreMicInput` is set to `false`, `Device.testPreflight(token, options)` API requires a [token](https://www.twilio.com/docs/iam/access-tokens) with a TwiML app that can record an audio from a microphone and the ability to play the recorded audio back to the browser. In order to achieve this, we need two TwiML endpoints: one to capture and record the audio, and another one to play the recorded audio.
 
 TwiML Bins
 ----------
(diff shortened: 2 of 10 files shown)


ryan-rowland:
Posting here for the record. @charliesantos and I discussed a potential solution that will allow this override in all browsers in slack. The thinking is `AudioContext.createAudioElementSource()` -> `AudioContext.createMediaStreamDestination()`. Something like the following:
```js
const audioContext = Device._audioContext;
const src = audioContext.createAudioElementSource(el);
const dest = audioContext.createMediaStreamDestination();
src.connect(dest);
connection._setInputDeviceStream(dest.stream);
el.play();
```

PREFLIGHT.md
```diff
@@ -20,6 +20,7 @@ The `options` parameter is a JavaScript object containing configuration settings
 | `codecPreferences` | `['pcmu', 'opus']` | An ordered list of preferred codecs. |
 | `debug` | `false` | Can be `true` or `false`. Set this property to true to enable debug logging in your browser console. |
 | `edge` | `roaming` | Specifies which Twilio `edge` to use when initiating the test call. Please see documentation on [edges](https://www.twilio.com/docs/voice/client/edges). |
+| `ignoreMicInput` | `false` | If set to `true`, the test call will ignore microphone input and will use a default audio file. If set to `false`, the test call will capture the audio from the microphone. Setting this to `true` is only supported on Chrome and will throw an `InputStreamNotSupportedError` on other browsers |
```
ryan-rowland, line 23: WDYT about `fakeInputAudio`?

lib/twilio/device.ts
```diff
@@ -103,6 +103,11 @@ export interface IExtendedDeviceOptions extends Device.Options {
    */
   ignoreBrowserSupport?: boolean;
 
+  /**
+   * Input stream to use instead of reading from mic
+   */
+  inputStream?: MediaStream;
```
ryan-rowland, line 109: fakeInputStream?

lib/twilio/preflight/preflight.ts
```diff
@@ -248,10 +251,78 @@ export class PreflightTest extends EventEmitter {
...
+  private _getStreamFromFile(): Promise<MediaStream> {
+    return new Promise((resolve, reject) => {
+      if (typeof Audio.prototype.captureStream !== 'function') {
+        return reject(new InputStreamNotSupportedError(
+          'Getting a MediaStream from an audio file is not supported by this browser.' +
+          'Please set PreflightTest.Options.ignoreMicInput to false',
+        ));
+      }
+
+      const url = `${C.SOUNDS_BASE_URL}/cowbell.mp3?cache=${C.RELEASE_VERSION}`;
```
ryan-rowland, line 266: Should this be customizable?

PREFLIGHT.md
charliesantos: The input audio is not really fake, I think. Because it's coming from a file. Maybe fakeMicInput?

lib/twilio/preflight/preflight.ts
```diff
@@ -475,6 +572,11 @@ export namespace PreflightTest {
      * Device class to use.
      */
     deviceFactory?: new (token: string, options: Device.Options) => Device;
+
+    /**
+     * Input stream to use instead of reading from mic
+     */
+    inputStream?: MediaStream;
```
ryan-rowland, line 579: fakeInputStream?

lib/twilio/device.ts
charliesantos: The input stream here is not really fake, I think. Because it's coming from a file. Maybe fileInputStream?

lib/twilio/preflight/preflight.ts
charliesantos: I don't think it should be customizable. We need to play an audio that doesn't have gaps to easily analyze non-zero levels. We don't want customer providing bad audio files. Also, this is played muted, the customer won't hear this.

PREFLIGHT.md
ryan-rowland: I don't think we specify "mic" anywhere else, really. We refer to it as input audio. But I agree fakeInputAudio isn't totally correct. WDYT about:
```ts
/**
  * False does nothing, uses input audio from device
  * True uses cowbell
  * String references a URI (url or data blob) to play a custom sound
  */
inputAudioFromFile: boolean | string = false;
```

lib/twilio/device.ts
ryan-rowland: `inputStreamFromFile`? Can this be null? Or are we setting this to mic audio if the option isn't passed? In that case `inputStream` is fine.

lib/twilio/preflight/preflight.ts
ryan-rowland: Are we making this part of the public Device.Options interface?

If so, I think it should be customizable in those options. If not, then we can ignore.

Also, as a side note, keep in mind the issues we've had in QE with polling every second using cowbell because there are periods of silence between dings!

lib/twilio/device.ts
charliesantos: Yes, this can be null. Device uses the mic audio if this option isn't passed.

PREFLIGHT.md
ryan-rowland: ^^ Assuming we go with customizable. Otherwise `fakeMicInput: boolean` would work if this is a private, undocumented interface in Device.Options.

lib/twilio/device.ts
ryan-rowland: 👍 I'm fine with either `fileInputStream` or `inputStreamFromFile` then

lib/twilio/preflight/preflight.ts
charliesantos: No, this is not part of the public Device.Options interface, so I think it's ok to make it not customizable.

PREFLIGHT.md
charliesantos: Cool. Yeah this is a private , undocumented interface in Device.Options.

PREFLIGHT_TWIML.md
```diff
@@ -44,3 +44,19 @@ Creating the TwiML App
 Now that we have created our TwiML Bins, let's create our TwiML app by going to the [TwiML Apps](https://www.twilio.com/console/voice/twiml/apps) page. Click the plus button on that screen and enter a friendly name that you prefer. Under Voice request url, enter the **TwiML Bin's Record** url that you created in the previous section, and then click the **Create** button.
 
 On that same page, open the TwiML app that you just created by clicking on it and make note of the `SID`. You can now use this TwiML app to generate your [token](https://www.twilio.com/docs/iam/access-tokens) when calling `Device.testPreflight(token, options)` API.
+
+
+TwiML App - Echo
+================
+
+If `PreflightTest.Options.ignoreMicInput` is set to `true`, `Device.testPreflight(token, options)` API requires a [token](https://www.twilio.com/docs/iam/access-tokens) with a single TwiML app that can capture and play an audio. Following the [previous steps](#twiml-bins), create a TwiML Bin using the following template, and name it `Echo`.
```
ryan-rowland, line 52: I thought we were going to change this to "fakeMicInput"? I think I resolved the comment because my other concerns were resolved, but maybe we lost track of this. My bad.

lib/twilio/device.ts
```diff
@@ -242,6 +247,20 @@ class Device extends EventEmitter {
    */
   private static _isUnifiedPlanDefault: boolean | undefined;
 
+  /**
+   * Initialize the AudioContext for the {@link Device} class and returns it
```
ryan-rowland, line 251: > Initializes the AudioContext instance shared across the Client SDK, or returns the existing instance if one has already been initialized.

ryan-rowland: Also, we might want to call this `_getOrCreateAudioContext()`

lib/twilio/preflight/preflight.ts
```diff
@@ -7,11 +7,16 @@
...
 import Device from '../device';
+import { NotSupportedError } from '../errors';
 import { RTCSampleTotals } from '../rtc/sample';
 import RTCSample from '../rtc/sample';
 import RTCWarning from '../rtc/warning';
+import StatsMonitor from '../statsMonitor';
 import { NetworkTiming, TimeMeasurement } from './timing';
 
+const C = require('../constants');
+const ECHO_TEST_DURATION = 20000;
```
ryan-rowland, line 18: Seems funny to declare a constants after importing a constants file. Should this be declared in that file?

lib/twilio/preflight/preflight.ts
```diff
@@ -248,10 +249,95 @@ export class PreflightTest extends EventEmitter {
     }, {} as any);
   }
 
+  /**
+   * Returns a MediaStream from a media file
+   */
+  private _getStreamFromFile(): MediaStream {
+    const audioContext = ((this._options.deviceFactory || Device) as any)['_initAudioContext']();
```
ryan-rowland, line 256: We shouldn't be calling private fields like this, this is cheating TS 😛 

Maybe we accept a `getOrCreateAudioContext` method argument on PreflightTest construction. We can pass in the method from Device.

lib/twilio/preflight/preflight.ts
```diff
@@ -248,10 +249,95 @@ export class PreflightTest extends EventEmitter {
...
+    if (!audioContext) {
+      throw new NotSupportedError('AudioContext is not supported by this browser.');
```
ryan-rowland, line 258: We should be more specific about where this error is coming from. Like "Can't fake input audio stream: AudioContext is not supported by this browser."

lib/twilio/preflight/preflight.ts
```diff
@@ -248,10 +249,95 @@ export class PreflightTest extends EventEmitter {
...
+    }
+
+    const url = `${C.SOUNDS_BASE_URL}/cowbell.mp3?cache=${C.RELEASE_VERSION}`;
```
ryan-rowland, line 261: Constants file?

lib/twilio/preflight/preflight.ts
```diff
@@ -248,10 +249,95 @@ export class PreflightTest extends EventEmitter {
...
+    try {
+      this._device = new (options.deviceFactory || Device)(token, {
+        codecPreferences: options.codecPreferences,
+        debug: options.debug,
+        edge: options.edge,
+        fileInputStream: options.fileInputStream,
+        // Silence the sounds if we're not using the mic
+        sounds: options.ignoreMicInput ? {
+          disconnect: 'http://empty.mp3',
+          outgoing: 'http://empty.mp3',
```
ryan-rowland, line 289: ?? Won't this throw an error / warning in console?

PREFLIGHT_TWIML.md
charliesantos: Probably misunderstanding on my side. I'll update to fakeMicInput

lib/twilio/preflight/preflight.ts
charliesantos: I'll move it to constants file.

lib/twilio/preflight/preflight.ts
charliesantos: Yes, but should be benign.

ryan-rowland: We get complaints when we throw benign errors / warnings. They can be red herrings for developers when they're trying to debug. Can't we pass `disconnect: false` or `disconnect: null` to not play the sound?

charliesantos: Passing false or null will end up playing the default sounds. I remember this is also what we've suggested to our customers for muting the sounds. But I can understand the concern. I'll try to think of a better way.

ryan-rowland: I'm fairly certain passing `false` or `null` used to work this way. Looking through the code now I can see that was changed at some point:
https://github.com/twilio/twilio-client.js/blob/master/lib/twilio/device.ts#L672

Notice the last argument, `value`, that isn't ever passed anymore. You can also see `enabledSounds` which also seems to no longer be updated correctly.

charliesantos: Went all the way to 1.4 and passing false/null doesn't seem to work 😞 
if I'm understanding the code correctly, enabledSounds seems to be for `device.sounds` API which can be used to mute sounds - `device.sounds.incoming(false)`.

Looking at the docs, it seems like that is now deprecated and moved to `device.audio`. I updated to use `device.audio` APIs to mute sounds. Please check :)

PR #170, opened 2020-05-19
[CLIENT-7502] Emit samples even when MOS is `null`.

mhuynh5757:
**Contributing to Twilio**

### JIRA link(s):

- [CLIENT-7502](https://issues.corp.twilio.com/browse/CLIENT-7502)

### Description

This PR introduces a change to Preflight Test such that samples emitted by the connection are also emitted by the test even when MOS is `null`.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/PREFLIGHT_TWIML.md
+++ b/PREFLIGHT_TWIML.md
@@ -17,6 +17,8 @@ Create a new TwiML Bin with the plus button on that screen and use `Playback` as
 <Response>
   <Say>You said:</Say>
   <Play loop="1">{{RecordingUrl}}</Play>
+  <Say>Now waiting for a few seconds to gather audio performance metrics.</Say>
+  <Pause length="3">
   <Say>Hanging up now.</Say>
 </Response>
 ```
--- a/PREFLIGHT_TWIML.md
+++ b/PREFLIGHT_TWIML.md
@@ -31,7 +31,7 @@ Using the [TwiML Bin](https://www.twilio.com/console/twiml-bins) page, let's cre
 
 <Response>
   <Say>Record a message in 3, 2, 1</Say>
-  <Record maxLength="3" action="https://my-record-twiml-url"></Record>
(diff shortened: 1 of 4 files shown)


PREFLIGHT_TWIML.md
```diff
@@ -31,7 +31,7 @@ Using the [TwiML Bin](https://www.twilio.com/console/twiml-bins) page, let's cre
 
 <Response>
   <Say>Record a message in 3, 2, 1</Say>
-  <Record maxLength="3" action="https://my-record-twiml-url"></Record>
+  <Record maxLength="5" action="https://my-record-twiml-url"></Record>
```
charliesantos, line 34: Testing this more, it doesn't seem to solve the problem and it only feels weird because we now have a longer pause. I think we need to lengthen the TwiML playback, same approach we did with network test tool. That is, let's put this back to 3 seconds and update the `TwiML App -Playback` template to the following:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<Response>
  <Say>You said:</Say>
  <Play loop="1">{{RecordingUrl}}</Play>
  <Say>Now waiting fifteen seconds to gather audio performance metrics.</Say>
  <Pause length="15"/>
  <Say>Hanging up now.</Say>
</Response>
```

charliesantos: Actually, waiting 15 seconds might trigger constant audio level warnings. Let's keep the maxlength of 5 seconds in your PR, and update the `TwiML App - Playback` to this instead. What do you think?
```xml
<?xml version="1.0" encoding="UTF-8"?>
<Response>
  <Say>You said:</Say>
  <Play loop="1">{{RecordingUrl}}</Play>
  <Say>Now waiting a few seconds to gather audio performance metrics.</Say>
  <Pause length="3"/>
  <Say>Hanging up now.</Say>
</Response>
```

Issue #172, opened 2020-05-24
[FEATURE] Add docs on how to use modern import


mesqueeb:
currently the docs only say:
```ts
const Device = require('twilio-client').Device;
```
But since this is for usage in a browser, and most browser apps now-a-days use modern import/export syntax, I'd love to know which one is correct:
```ts
import twilioClient from 'twilio-client'
const Device = twilioClient.Device
```
or
```ts
import * as twilioClient from 'twilio-client'
const Device = twilioClient.Device
```
or
```ts
import { Device } from 'twilio-client'
```

PS: I use typescript

charliesantos:
Hi @mesqueeb ,

Thank you for this suggestion. This has been addressed in this PR https://github.com/twilio/twilio-client.js/pull/173

PR #173, opened 2020-05-26
GH 172 | Adding import to readme

charliesantos:
https://github.com/twilio/twilio-client.js/issues/172

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,14 +25,14 @@ We recommend using `npm` to add the Client SDK as a dependency.
 npm install twilio-client --save
 ```
 
-Using this method, you can `import` twilio-client.js like so:
+Using this method, you can `import` twilio-client.js using ES Module or TypeScript syntax:
 
 ```js
 import { Device } from 'twilio-client';
 
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,7 +25,14 @@ We recommend using `npm` to add the Client SDK as a dependency.
...
 
-Using this method, you can `require` twilio-client.js like so:
+Using this method, you can `import` twilio-client.js like so:
+
+```js
+import { Device } from 'twilio-client';
+
+```
+
+Or using `require`:
```
ryan-rowland, line 35: When would a customer use `import` and when would they use `require`? We should clarify why there are two ways to do this.

charliesantos: What do you think about removing the section about `require`? I think if we leave just the `import` section, it's pretty much straight forward.

ryan-rowland: We can probably just say something like "In ES6 and newer, and in TypeScript..." and "In older JS syntax..." and it should be clear enough

charliesantos: I updated the readme, please check again.

PR #184, opened 2020-06-06
Handle scenario where no candidates are selected on preflight.

charliesantos:
Handle scenario where no candidates are selected during ICE negotiation on preflight.

### DIFF
--- a/PREFLIGHT.md
+++ b/PREFLIGHT.md
@@ -108,7 +108,7 @@ Raised when `PreflightTest.status` has transitioned to `PreflightTest.Status.Com
   /**
    * Whether a TURN server is required to connect to media.
    * This is dependent on your ICE server configuration and is set to true if the selected ICE candidate is of type `relay`.
-   * Also, this is undefined if there's no candidate selected during ICE negotiation.
+   * Also, this is undefined if there are no candidates selected during ICE negotiation.
    * See `PreflightTest.Options.iceServers` for more details.
    */
   "isTurnRequired": false,
--- a/PREFLIGHT.md
+++ b/PREFLIGHT.md
@@ -107,8 +107,8 @@ Raised when `PreflightTest.status` has transitioned to `PreflightTest.Status.Com
 
   /**
    * Whether a TURN server is required to connect to media.
-   * This is dependent on your ICE server configuration and is set to true if the selected ICE candidate is of type `relay`.
-   * Also, this is undefined if there are no candidates selected during ICE negotiation.
+   * This is dependent on the selected ICE candidates, and will be true if either is of type "relay",
+   * false if both are of another type, or undefined if there are no selected ICE candidates.
    * See `PreflightTest.Options.iceServers` for more details.
    */
   "isTurnRequired": false,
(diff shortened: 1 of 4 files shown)


PREFLIGHT.md
```diff
@@ -107,7 +107,8 @@ Raised when `PreflightTest.status` has transitioned to `PreflightTest.Status.Com
 
   /**
    * Whether a TURN server is required to connect to media.
-   * This is dependent on your ICE server configuration and is set to true if the selected ICE candidate is of type `relay`
+   * This is dependent on your ICE server configuration and is set to true if the selected ICE candidate is of type `relay`.
+   * Also, this is undefined if there are no candidates selected during ICE negotiation.
```
ryan-rowland, line 111: > This is dependent on the selected ICE candidates, and will be true if either is of type "relay", false if both are of another type, or undefined if there are no selected ICE candidates.

lib/twilio/preflight/preflight.ts
```diff
@@ -900,11 +903,12 @@ export namespace PreflightTest {
     iceCandidates: RTCIceCandidateStats[];
 
     /**
-     * Whether a TURN server is required to connect to media.
+     * A boolean value indicating whether a TURN server is required to connect to media.
+     * This is undefined if there's no candidate selected during ICE negotiation.
```
ryan-rowland, line 907: See suggestion above

lib/twilio/preflight/preflight.ts
```diff
@@ -923,8 +927,9 @@ export namespace PreflightTest {
 
     /**
      * The ICE candidate pair used to connect to media.
+     * This is null if there's no candidate selected during ICE negotiation.
      */
-    selectedIceCandidatePair: RTCSelectedIceCandidatePair;
+    selectedIceCandidatePair?: RTCSelectedIceCandidatePair;
```
ryan-rowland, line 932: Undefined or null? Looks like undefined, but let's be consistent with other fields -- are other fields null or undefined in this scenario? If null, we need to specify `RTCSelectedIceCandidatePair | null` instead of the question mark syntax.

In both places I think we can simplify to:
```
The ICE candidate pair used to connect to media, if candidates were selected.
```

charliesantos: It is undefined. Other fields are also undefined. Not sure why I put null :)

lib/twilio/preflight/preflight.ts
```diff
@@ -718,6 +720,7 @@ export namespace PreflightTest {
 
     /**
      * The ICE candidate pair used to connect to media.
+     * This is null if there's no candidate selected during ICE negotiation.
```
ryan-rowland, line 723: Missed a null reference here

lib/twilio/rtc/stats.js
```diff
@@ -98,10 +98,10 @@ function getRTCIceCandidates(peerConnection) {
     // Build the return object
     return {
       iceCandidates: [...localCandidates, ...remoteCandidates],
-      selectedIceCandidatePair: {
+      selectedIceCandidatePair: selectedCandidatePairReport ? {
         localCandidate: localCandidates.find(candidate => candidate.id === selectedCandidatePairReport.localCandidateId),
         remoteCandidate: remoteCandidates.find(candidate => candidate.id === selectedCandidatePairReport.remoteCandidateId),
-      },
+      } : null,
```
ryan-rowland, line 104: And here -- this should be undefined, right?

PR #196, opened 2020-06-30
CLIENT-7786 | [Preflight] Fix output not muting for non default device

charliesantos:
**Contributing to Twilio**

### JIRA link(s):

- [CLIENT-7786](https://issues.corp.twilio.com/browse/CLIENT-7786)

### Description

This PR fixes an issue where master audio is not muted after changing device. This PR is now making sure all output audio are muted whenever clientjs updates the output.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/lib/twilio/preflight/preflight.ts
+++ b/lib/twilio/preflight/preflight.ts
@@ -17,6 +17,17 @@ import { NetworkTiming, TimeMeasurement } from './timing';
 
 const { COWBELL_AUDIO_URL, ECHO_TEST_DURATION } = require('../constants');
 
+/**
+ * Represents the audio output object coming from Client SDK's PeerConnection object.
+ * @internalapi
+ */
+export interface AudioOutput {
+  /**
+   * The audio element used to play out the sound.
+   */
+  audio: HTMLAudioElement;
+}
+
 export declare interface PreflightTest {
   /**
    * Raised when [[PreflightTest.status]] has transitioned to [[PreflightTest.Status.Completed]].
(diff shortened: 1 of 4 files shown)


lib/twilio/preflight/preflight.ts
```diff
@@ -17,6 +17,17 @@ import { NetworkTiming, TimeMeasurement } from './timing';
 
 const { COWBELL_AUDIO_URL, ECHO_TEST_DURATION } = require('../constants');
 
+/**
+ * Represents the audio output object coming from Client SDK's PeerConnection object.
+ * @internalapi
+ */
+export interface AudioOutput {
```
ryan-rowland, line 24: Should this be AudioElement?

lib/twilio/preflight/preflight.ts
```diff
@@ -359,20 +370,26 @@ export class PreflightTest extends EventEmitter {
...
-        stream._masterAudio.muted = true;
-      }
-      stream.__fallbackOnAddTrack = stream._fallbackOnAddTrack;
-      stream._fallbackOnAddTrack = (...args: any[]) => {
-        const rval = stream.__fallbackOnAddTrack(...args);
-        stream.outputs.get('default').audio.muted = true;
-        return rval;
-      };
+      const muteOutputs = (outputs: Map<string, AudioOutput>) =>
+        outputs.forEach((output: AudioOutput) => output.audio.muted = true);
```
ryan-rowland, line 379: Looks like we're modifying by reference here. We should either:
- Modify by reference and not assign `muteOutputs` or
- Not modify by reference, and `muteOutputs` would be a modified copy.

It's a bit confusing to modify by reference and create a new variable.

lib/twilio/preflight/preflight.ts
```diff
@@ -359,20 +370,26 @@ export class PreflightTest extends EventEmitter {
...
+
+      ['onAddTrack', 'fallbackOnAddTrack', 'updateAudioOutputs'].forEach(handler => {
+        stream[`_orig${handler}`] = stream[`_${handler}`];
+        stream[`_${handler}`] = (...args: any[]) => {
```
ryan-rowland, line 383: We shouldn't be accessing internal members of child classes.

lib/twilio/preflight/preflight.ts
charliesantos: It's an object that contains an Audio Element..

lib/twilio/preflight/preflight.ts
charliesantos: I think that is the intention. We want a function that mutes a list of audio out objects.

lib/twilio/preflight/preflight.ts
charliesantos: Yeah, accessing the internals is not ideal. But since this is for preflight, and we don't want to expose a public method just for it, I was thinking it's probably ok. Let me think about it more and see if there's a better way.

charliesantos: I updated so that we don't have to access these internal handlers.

PR #200, opened 2020-07-23
CLIENT-7725 | Add insights-connection-error warning

ryan-rowland:
### JIRA link(s):

- [CLIENT-7725](https://issues.corp.twilio.com/browse/CLIENT-7725)

#### Breaking Changes

* (Breaking Change) We now emit `(warning: PreflightTest.Warning)` object from PreflightTest.on('warning'),
  rather than `(name: string, data: RTCWarning)`. The `PreflightTest.Warning` object has been updated
  to match the following interface:
  ```ts
  export interface Warning {
    description: string;
    name: string;
    rtcWarning?: RTCWarning;
  }
  ```

#### Features

* We now emit a PreflightTest.Warning (`insights-connection-error`) the first time Insights emits an
  error, and add that Warning in `Report.warnings`.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


lib/twilio/preflight/preflight.ts
```diff
@@ -944,6 +947,24 @@ export namespace PreflightTest {
     /**
      * List of warning names and warning data detected during this test.
      */
-    warnings: Warning[];
+    warnings: PreflightTest.Warning[];
+  }
+
+  /**
+   * A warning that can be raised by Preflight, and returned in the Report.warnings field.
```
charliesantos, line 954:
```suggestion
   * A warning that can be raised by Preflight, and returned in the [[PreflightTest.Report.warnings]] field.
```

PREFLIGHT.md
```diff
@@ -231,7 +231,7 @@ Raised when `PreflightTest.status` has transitioned to `PreflightTest.Status.Com
 
   /**
    * Warnings detected during the test.
-   * These are coming from Connection.on('warning').
+   * This will emit any RTCWarnings or PreflightTest.Warnings encountered
```
charliesantos, line 234: With this change, we don't emit RTCWarnings directly anymore right (warnings coming directly from Connection)? Maybe this can be simplified like the following?
```
  /**
   * PreflightTest.Warnings detected during the test.
   */
```

PREFLIGHT.md
```diff
@@ -247,7 +247,7 @@ Raised when `PreflightTest.status` has transitioned to `PreflightTest.Status.Fai
 This event is published every second and is raised when the [Connection](https://www.twilio.com/docs/voice/client/javascript/connection) gets a webrtc sample object. The `sample` object is coming from [Connection.on('sample')](https://www.twilio.com/docs/voice/client/javascript/connection#sample) and uses the same `sample` format.
 
 #### .on('warning', handler(warningName, warningData))
-Raised whenever the [Connection](https://www.twilio.com/docs/voice/client/javascript/connection) encounters a warning. This is coming from [Connection.on('warning')](https://www.twilio.com/docs/voice/client/javascript/connection#onwarning-handlerwarningname) and uses the same API format.
+Raised whenever the [Connection](https://www.twilio.com/docs/voice/client/javascript/connection) encounters an RTCWarning, or Preflight encounters a warning. RTCWarnings come from [Connection.on('warning')](https://www.twilio.com/docs/voice/client/javascript/connection#onwarning-handlerwarningname) and uses the same API format, and `PreflightTest.Warning`s are warnings generated by preflight test.
```
charliesantos, line 250: With this change, we don't emit RTCWarnings directly anymore right (warnings coming directly from Connection) and we instead wrap it with `PreflightTest.Warning`? It's probably helpful if we provide an example here so the user understands the object format. We should also update `#### .on('warning', handler(warningName, warningData))` to something like `#### .on('warning', handler(warning))` to match the newly implemented API.
What do you think about this?

#### .on('warning', handler(warning))
Raised when the test encounters a warning. Example warning data:

```js
{
  /**
   * Name of the warning.
   * See https://www.twilio.com/docs/voice/insights/call-quality-events-twilio-client-sdk
   */
  name: 'insights-connection-error',

  /**
   * Description of the Warning.
   */
  description: 'Received an error when attempting to connect to Insights gateway',

  /**
   * Optional RTCWarning data coming from Connection.on('warning')
   * See https://www.twilio.com/docs/voice/client/javascript/connection#onwarning-handlerwarningname
   */
  rtcWarning: {...}
}
```

PREFLIGHT.md
```diff
@@ -247,7 +246,28 @@ Raised when `PreflightTest.status` has transitioned to `PreflightTest.Status.Fai
 This event is published every second and is raised when the [Connection](https://www.twilio.com/docs/voice/client/javascript/connection) gets a webrtc sample object. The `sample` object is coming from [Connection.on('sample')](https://www.twilio.com/docs/voice/client/javascript/connection#sample) and uses the same `sample` format.
 
 #### .on('warning', handler(warningName, warningData))
```
charliesantos, line 248:
```suggestion
#### .on('warning', handler(warning))
```

PR #213, opened 2020-08-25
[CLIENT-7613] Update `constant-audio-level` warning.

mhuynh5757:
**Contributing to Twilio**

### JIRA link(s):

- [CLIENT-7613](https://issues.corp.twilio.com/browse/CLIENT-7613)

### Description

This PR updates the `constant-audio-level` warning threshold to match the IDL as per CLIENT-7538.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,9 @@
 
 Changes
 -------
+
 * The behavior for raising the `high-packet-loss` warning has been updated. Now, the most recent seven seconds of packet-loss samples are recorded and analyzed. If the average of these samples is greater than `3%` packet-loss, then the warning is raised. When the average packet-loss over the most recent seven seconds is less than or equal to `1%` packet-loss, then the warning is cleared.
+* The behavior for raising the `constant-audio-level` warning has been updated. Now, the most recent ten seconds of volume values are recorded and then analyzed. If the standard deviation of these samples is less than 1% of the maximum audio value, then the warning is raised. When the standard deviation is greater than 1% and the warning has already been raised, then the warning is cleared.
 
 1.12.5 (Sept 22, 2020)
 ====================
@@ -15,7 +17,7 @@ Bug Fixes
 * Fixed an issue where the disconnect sound plays after the caller cancelled the incoming call.
 
 1.12.4 (Sept 4, 2020)
-====================
+=====================
 
 Bug Fixes
 ---------
@@ -35,7 +37,6 @@ Changes
 
 * We now log selected [ICE candidate pair](https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidatePair) to Insights. This will help with isolating issues should they arise.
 * Fixed an issue where the transportClose event listener was not being cleaned up appropriately when a Connection is closed, causing MaxListenersExceededWarning on the console.
-* The behavior for raising the `constant-audio-level` warning has been updated. Now, the most recent ten seconds of volume values are recorded and then analyzed. If the standard deviation of these samples is less than 1% of the maximum audio value, then the warning is raised. When the standard deviation is greater than 1% and the warning has already been raised, then the warning is cleared.
 
 1.12.2 (August 6, 2020)
 =======================
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ Changes
 
 * We now log selected [ICE candidate pair](https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidatePair) to Insights. This will help with isolating issues should they arise.
 * Fixed an issue where the transportClose event listener was not being cleaned up appropriately when a Connection is closed, causing MaxListenersExceededWarning on the console.
+* The behavior for raising the `constant-audio-level` warning has been updated. Now, the most recent ten seconds of volume values are recorded and then analyzed. If the standard deviation of these samples is less than 1% of the maximum audio value, then the warning is raised. When the standard deviation is greater than 1% and the warning has already been raised, then the warning is cleared.
 
 1.12.2 (August 6, 2020)
 =======================
--- a/lib/twilio/statsMonitor.ts
+++ b/lib/twilio/statsMonitor.ts
@@ -77,21 +77,6 @@ function countLow(min: number, values: number[]): number {
 }
 
 /**
- * Average the set of values.
- * @private
- * @param values - The values to average.
- * @returns The average of those values or `null` if the set of values is empty.
- */
-function averageValues(values: number[]): number | null {
-  return values.length
-    ? values.reduce(
-        (partialSum: number, value: number) => partialSum + value,
-        0,
-      ) / values.length
-    : null;
-}
-
-/**
  * Calculate the standard deviation from a list of numbers.
  * @private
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ Changes
 
 * We now log selected [ICE candidate pair](https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidatePair) to Insights. This will help with isolating issues should they arise.
 * Fixed an issue where the transportClose event listener was not being cleaned up appropriately when a Connection is closed, causing MaxListenersExceededWarning on the console.
+* The behavior for raising the `constant-audio-level` warning has been updated. Now, the most recent ten seconds of volume values are recorded and then analyzed. If the standard deviation of these samples is less than 1% of the maximum audio value, then the warning is raised. When the standard deviation is greater than 1% and the warning has already been raised, then the warning is cleared.
```
charliesantos, line 9: This entry should probably be under 1.12.4 (In Progress)

lib/twilio/statsMonitor.ts
```diff
@@ -13,7 +13,7 @@ const getRTCStats = require('./rtc/stats');
 const Mos = require('./rtc/mos');
 
 // How many samples we use when testing metric thresholds
-const SAMPLE_COUNT_METRICS = 5;
+const SAMPLE_COUNT_METRICS = 10;
```
charliesantos, line 16: Changing this to 10 affects how we calculate other metrics such as jitter, rtt, packets lost , and mos right? Since you are already adding `sampleCount` for audio levels in the `DEFAULT_THRESHOLDS`, we should probably keep this `SAMPLE_COUNT_METRICS` to 5.

lib/twilio/statsMonitor.ts
```diff
@@ -72,6 +72,41 @@ function countLow(min: number, values: number[]): number {
   return values.reduce((lowCount, value) => lowCount += (value < min) ? 1 : 0, 0);
 }
 
+/**
+ * Calculate the standard deviation from a list of numbers.
+ * @private
+ * @param values The list of numbers to calculate the standard deviation from.
```
charliesantos, line 78: `@param sampleValues `, to match the actual parameter name. Or rename `sampleValues` to `values`, which ever you prefer.

lib/twilio/statsMonitor.ts
```diff
@@ -72,6 +72,41 @@ function countLow(min: number, values: number[]): number {
...
+ * @returns The standard deviation of a list of numbers.
+ */
+function calculateStandardDeviation(sampleValues: number[]) {
+  const sampleAverage: number = sampleValues.reduce(
+    (partialSum: number, sample: number) => partialSum + sample,
+    0,
+  ) / sampleValues.length;
```
charliesantos, line 85: What happens if length is zero in the case of no samples gathered?

lib/twilio/statsMonitor.ts
```diff
@@ -434,19 +486,27 @@ class StatsMonitor extends EventEmitter {
...
-      this._currentStreaks.set(statName, streak);
+      const stdDev: number = calculateStandardDeviation(volumeSamples);
 
-      if (streak >= limits.maxDuration) {
-        this._raiseWarning(statName, 'maxDuration', { value: streak });
-      } else if (streak === 0) {
+      if (stdDev < 327.67) {
+        const currentStreak: number = prevStreak + 1;
+        this._currentStreaks.set(statName, currentStreak);
+        this._raiseWarning(statName, 'maxDuration', { value: currentStreak });
```
charliesantos, line 507: It doesn't make sense to call this warning `maxDuration`, but we are calculating min standard deviation. I suggest we keep the calculations for limits.maxDuration as is. Then add a new threshold and call it minStandardDeviation. That means, in `DEFAULT_THRESHOLDS`, you will have something like

```
audioInputLevel: { minStandardDeviation: .01, maxValue: 32767 },
...
```

Then inside this same `_raiseWarningsForStat` function, you will have a different else block.
```
if (typeof limits.minStandardDeviation === 'number') {
// Do your stuff here
}
```

lib/twilio/statsMonitor.ts
```diff
@@ -455,6 +526,33 @@ class StatsMonitor extends EventEmitter {
...
+        sampleSets.splice(0, sampleSets.length - limits.sampleCount);
+      }
+      const flatSamples: number[] = flattenSamples(sampleSets.slice(-sampleCount));
+      const stdDev: number | null = calculateStandardDeviation(flatSamples);
+
+      if (typeof stdDev !== 'number') {
+        return;
+      }
+
+      const prevStreak: number = this._currentStreaks.get(statName) || 0;
```
charliesantos, line 544: Why are we still using streaks? We don't need this right?

lib/twilio/statsMonitor.ts
```diff
@@ -455,6 +526,33 @@ class StatsMonitor extends EventEmitter {
...
+
+      if (stdDev < limits.minStandardDeviation) {
+        const currentStreak: number = prevStreak + 1;
+        this._currentStreaks.set(statName, currentStreak);
+        this._raiseWarning(statName, 'minStandardDeviation', { value: currentStreak });
```
charliesantos, line 549: Why emitting the streak value? It should be the value of the calculated standard deviation no?

lib/twilio/statsMonitor.ts
```diff
@@ -455,6 +526,33 @@ class StatsMonitor extends EventEmitter {
...
+      } else {
+        this._currentStreaks.set(statName, 0);
+        this._clearWarning(statName, 'minStandardDeviation', { value: prevStreak });
```
charliesantos, line 552: Why emitting the streak value? It should be the value of the calculated standard deviation no?

charliesantos (changes requested):
Left a few comments. Also found some issues:
1) After constant audio input level warning is raised, it gets cleared right away after a few seconds. Try setting the audio input level to always zero.
2) I cannot get the constant audio **_output_** level warning to raise. Try setting the output level to zero.

mhuynh5757:
From this line https://github.com/twilio/twilio-client.js/blob/master/lib/twilio/connection.ts#L1077 it seems like we don't want to emit `constant-audio-output-level` warnings, do we want to change that?

charliesantos (approved):

lib/twilio/statsMonitor.ts
```diff
@@ -76,6 +76,45 @@ function countLow(min: number, values: number[]): number {
...
 }
 
+/**
+ * Calculate the standard deviation from a list of numbers.
+ * @private
+ * @param values The list of numbers to calculate the standard deviation from.
+ * @returns The standard deviation of a list of numbers.
+ */
+function calculateStandardDeviation(values: number[]): number | null {
+  if (!(values.length > 0)) {
```
ryan-rowland, line 86: So in other words, `if (values.length <= 0)`? 😛

ryan-rowland (changes requested):
Found one line quirky but otherwise LGTM

ryan-rowland (approved):

Issue #164, opened 2020-05-05
[BUG] Call is not ended after navigation in Safari 13.1


ostap0207:
- [x] I have verified that the issue occurs with the latest twilio.js release and is not marked as a known issue in the [CHANGELOG.md](https://github.com/twilio/twilio-client.js/blob/master/CHANGELOG.md).
 - [x] I reviewed the [Common Issues](https://github.com/twilio/twilio-client.js/blob/master/COMMON_ISSUES.md) and open GitHub issues and verified that this report represents a potentially new issue.
 - [x] I verified that the [Quickstart](https://github.com/TwilioDevEd/client-quickstart-js) application works in my environment.
 - [x] I am not sharing any [Personally Identifiable Information (PII)](https://www.twilio.com/docs/glossary/what-is-personally-identifiable-information-pii)
       or sensitive account information (API keys, credentials, etc.) when reporting this issue.

**Code to reproduce the issue:**
1. Start the client-quickstart-js in Safari 13.1
2. Call your phone number
3. Add a link for navigation in the browser console (it does not matter where the page goes):
```
      var a = document.createElement('a');
      var linkText = document.createTextNode("navigate");
      a.appendChild(linkText);
      a.title = "navigate";
      a.href = "/next_page";
      document.body.appendChild(a);
```
4. Click on the link

**Expected behavior:**

Phone call ends

**Actual behavior:**

Phone call continues after navigation

Debug logs:
```
[Info] Setting up VSP (twilio.min.js, line 99)
[Info] WSTransport.open() called... (twilio.min.js, line 99)
[Info] Attempting to connect... (twilio.min.js, line 99)
[Info] Closing and cleaning up WebSocket... (twilio.min.js, line 99)
[Info] No WebSocket to clean up. (twilio.min.js, line 99)
[Info] WebSocket opened successfully. (twilio.min.js, line 99)
[Info] Setting token and publishing listen (twilio.min.js, line 99)
[Info] Stream is ready (twilio.min.js, line 99)
[Log] Calling ***... (quickstart.js, line 93)
[Info] signalingState is "have-local-offer" (twilio.min.js, line 99)
[Info] pc.iceGatheringState is "gathering" (twilio.min.js, line 99)
[Info] pc.connectionState is "connecting" (twilio.min.js, line 99)
[...]

PR #221, opened 2020-09-14
CLIENT-7633 | Call destroy on unload and pagehide

ryan-rowland:
This is intended to address. https://github.com/twilio/twilio-client.js/issues/164

### JIRA link(s):

- [CLIENT-7633](https://issues.corp.twilio.com/browse/CLIENT-7633)

### Description

This PR updated our navigation behavior to call destroy (rather than just disconnectAll) and does the same for both the `unload` and `pagehide` events.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/lib/twilio/device.ts
+++ b/lib/twilio/device.ts
@@ -470,7 +470,7 @@ class Device extends EventEmitter {
   /**
    * Destroy the {@link Device}, freeing references to be garbage collected.
    */
-  destroy(): void {
+  destroy = (): void => {
     this._disconnectAll();
     this._stopRegistrationTimer();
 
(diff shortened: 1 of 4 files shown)


lib/twilio/device.ts
```diff
@@ -470,7 +470,7 @@ class Device extends EventEmitter {
   /**
    * Destroy the {@link Device}, freeing references to be garbage collected.
    */
-  destroy(): void {
+  destroy = (): void => {
```
charliesantos, line 473: Just curious, why do we need to change it to this? The other public functions in this class doesn't use arrow functions.

charliesantos (changes requested):

mhuynh5757 (approved):

charliesantos (approved):

PR #226, opened 2020-10-02
Open stream after listeners are added

ostap0207:
**Contributing to Twilio**

### Description

Currently `error` listener is added after the connection is open.
So if there is a synchronous error during connection open (for example
exeption) then we don't receive this error in the error listener.

We found this after noticing that if websocket connection to Twilio is
blocked by CSP then on iOS Safari it's not being reported as error.

On Chrome this does not reproduce as it will not throw an error in this
case, but instead call WebSocket error listener.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/lib/twilio/device.ts
+++ b/lib/twilio/device.ts
@@ -1227,6 +1227,7 @@ class Device extends EventEmitter {
     this.stream.addListener('invite', this._onSignalingInvite);
     this.stream.addListener('offline', this._onSignalingOffline);
     this.stream.addListener('ready', this._onSignalingReady);
+    this.stream.open();
   }
 
   /**
--- a/lib/twilio/pstream.js
+++ b/lib/twilio/pstream.js
@@ -100,12 +100,15 @@ function PStream(token, uris, options) {
   this.transport.on('error', this._handleTransportError);
   this.transport.on('message', this._handleTransportMessage);
   this.transport.on('open', this._handleTransportOpen);
-  this.transport.open();
 
   return this;
 }
 
 util.inherits(PStream, EventEmitter);
+
+PStream.prototype.open = function() {
+  this.transport.open();
+};
 
 PStream.prototype._handleTransportClose = function() {
   this.emit('transportClose');
--- a/tests/unit/device.ts
+++ b/tests/unit/device.ts
@@ -529,14 +529,25 @@ describe('Device', function() {
       let device: any;
       let options: any;
       let pStreamFactory: any;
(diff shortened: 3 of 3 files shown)


ostap0207:
Will test this manually to ensure it works in a bit.

ostap0207:
Works manually, not quite sure why CI fails.

charliesantos:
Thanks for submitting @ostap0207 . I'll take a look and also check why CI is failing.

charliesantos:
@ostap0207 , can you please provide a reproduction steps, along with code snippets that can simulate the issue you're seeing?

ostap0207:
Here are the steps based on the client-quickstart-js
1. Add the following line to the head tag. It allows all the required URLs except `wss://*.twilio.com` to simulated websockets being blocked.
```
<meta http-equiv="Content-Security-Policy" content="default-src 'self'; font-src 'self' *.gstatic.com; style-src 'self' 'unsafe-inline' *.googleapis.com; script-src 'self' 'unsafe-inline' 'unsafe-eval' *.googleapis.com *.twiliocdn.com; connect-src 'self' https://*.twilio.com *.twil.io; img-src 'self'; media-src 'self'">
```

2. Modify `quickstart.js` to setup the device after the listeners are added.
3. Open the page in iOS Safari

Expected: `Twilio.Device Error: ` appears on the screen. Works if you open the page in Chrome.
Actual: The error does not appear as the `error` listener is not being called. 
In the console log there is an error `Refused to connect to wss://chunderw-vpc-gll.twilio.com/signal because it does not appear in the connect-src directive of the Content Security Policy.`.

Or use the project attached below:
[client-quickstart-js.zip](https://github.com/twilio/twilio-client.js/files/5329781/client-quickstart-js.zip)

ostap0207:
The error that's not being triggered is https://github.com/twilio/twilio-client.js/blob/6d73cf602fbc729b4df6e4ecc644a0ddde8fe1d9/lib/twilio/wstransport.ts#L298. This error happens synchronously, but the listeners in device are currently added only after connection is created.

charliesantos:
Thanks @ostap0207 . We implemented asyncEmit before in Connection object. I'm thinking if that is a better implementation. I filed an internal ticket to track this. I'll keep you updated.
https://issues.corp.twilio.com/browse/CLIENT-8118

ostap0207:
Any updates on this?

PR #235, opened 2020-11-06
1.13.0 changelog and rename testPreflight to runPreflight

charliesantos:
Changelog for 1.13.0. I also added documentation to wagtail. See drafts below. They will be live once we get green light from QE.

PreflightTest class: https://www.twilio.com/docs/admin/pages/22197/view_draft/
Device.testPreflight: https://www.twilio.com/docs/admin/pages/1773/view_draft/#testpreflight
Connection.on('warning'): https://www.twilio.com/docs/admin/pages/1772/view_draft/#onwarning-handlerwarningname-warningdata

Edit: Also renamed testPreflight to runPreflight base on discussions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,28 @@
 1.13.0 (In Progress)
 ====================
 
+1.13.0-beta2 has been promoted to 1.13.0 GA. Here's what's new in 1.13.0.
+
+New Features
+------------
+
+### Voice diagnostics using PreflightTest API
+
+The SDK now supports a preflight test API which can help determine Voice calling readiness. The API creates a test call and will provide information to help troubleshoot call related issues. This new API is a static member of the [Device](https://www.twilio.com/docs/voice/client/javascript/device#twilio-device) class and can be used like the example below. Please see [Device.testPreflight](https://www.twilio.com/docs/voice/client/javascript/device#testpreflight) method and [PreflightTest](https://www.twilio.com/docs/voice/client/javascript/preflighttest) class for more details about this new API.
+
+```js
+// Initiate the test
+const preflight = Device.testPreflight(token, options);
+
+// Subscribe to events
+preflight.on('completed', (report) => console.log(report));
+preflight.on('failed', (error) => console.log(error));
+```
+
 Changes
 -------
+
+* [Connection.on('warning')](https://www.twilio.com/docs/voice/client/javascript/connection#onwarning-handlerwarningname) now provides data associated with the warning. This data can provide more details about the warning such as thresholds and WebRTC samples collected that caused the warning. The example below is a warning for high jitter. Please see [Voice Insights SDK Events Reference](https://www.twilio.com/docs/voice/insights/call-quality-events-twilio-client-sdk#warning-events) for a list of possible warnings.
 
 * Added `high-packets-lost-fraction` [network warning](https://www.twilio.com/docs/voice/insights/call-quality-events-twilio-client-sdk#network-warnings). This new warning is raised when the average of the most recent seven seconds of packet-loss samples is greater than `3%`. When the average packet-loss over the most recent seven seconds is less than or equal to `1%`, then the warning is cleared.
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,9 +1,66 @@
...
+New Features
+------------
+
+### Voice diagnostics using PreflightTest API
+
+The SDK now supports a preflight test API which can help determine Voice calling readiness. The API creates a test call and will provide information to help troubleshoot call related issues. This new API is a static member of the [Device](https://www.twilio.com/docs/voice/client/javascript/device#twilio-device) class and can be used like the example below. Please see [Device.testPreflight](https://www.twilio.com/docs/voice/client/javascript/device#testpreflight) method and [PreflightTest](https://www.twilio.com/docs/voice/client/javascript/preflighttest) class for more details about this new API.
+
+```js
+// Initiate the test
+const preflight = Device.testPreflight(token, options);
```
idelgado, line 15: Didn't think of this when I read the docs originally but any concerns about having the word `test` in the method name or object name? The reason I ask is in some languages/IDEs that string has special meaning and could be misinterpreted. Though API wise it would still be beneficial to have a noun rather than just an adjective, an alternative could be `preflightCheck`.

CHANGELOG.md
```diff
@@ -1,9 +1,66 @@
...
+
+// Subscribe to events
+preflight.on('completed', (report) => console.log(report));
+preflight.on('failed', (error) => console.log(error));
+```
```
idelgado, line 20: Other than this code snippet is there something more demonstrable we could reference? Maybe the app?

idelgado (changes requested):

CHANGELOG.md
idelgado: I would expect the call to action to be, go try using this "quick deploy"...

CHANGELOG.md
charliesantos: I haven't come across IDEs that complains `test` is in the method or object name. Can you provide an example? I don't think this is a problem on JS. Will this be a problem on mobile?

Regarding API name, API wise, I think the current convention (testFoo) makes sense. We usually name methods starting with a verb since it indicates that it will do some action. Noun is usually for class names no?

CHANGELOG.md
charliesantos: Good point on the quickstart app. I will link it in the changelog.

CHANGELOG.md
charliesantos: I think we should use `verbNoun` convention. I was just talking to @ryan-rowland , what do you think about `Device.runPreflightTest`?

charliesantos: If we end up renaming this, I have a little concern on RTC Diagnostics where we named all our tests using this convention. And that is already GA :(
```
testAudioInputDevice
testAudioOutputDevice
testMediaConnectionBitrate
testVideoInputDevice
```

CHANGELOG.md
charliesantos: I added a link to the API docs and quick deploy app. The API doc link is not live yet, but you can view the draft here. https://www.twilio.com/docs/admin/pages/22197/view_draft/

CHANGELOG.md
idelgado: How about just `runPreflight`? It's not actually a test as in it does not use a test unit framework. Might be better to use a more general naming convention for the interface.

CHANGELOG.md
```diff
@@ -1,16 +1,67 @@
 1.13.0 (In Progress)
 ====================
 
+1.13.0-beta2 has been promoted to 1.13.0 GA. Here's a summary of what is new in 1.13.0.
+
+New Features
+------------
+
+### Voice diagnostics using PreflightTest API
```
idelgado, line 9: > using runPreflight API

idelgado (approved):
Please make sure to update the IDL in addition to making the change here.

ryan-rowland (approved):

PR #237, opened 2020-11-14
Use latest version of Client.js

swarnava:


swarnava:
cc @ryan-rowland @charliesantos

PR #267, opened 2021-03-31
[CLIENT-8337] and [CLIENT-8338] Lazy Signaling and Device States

mhuynh5757:
**Contributing to Twilio**

### JIRA link(s):

- [CLIENT-8337](https://issues.corp.twilio.com/browse/CLIENT-8337)
- [CLIENT-8338](https://issues.corp.twilio.com/browse/CLIENT-8338)

### Description

This PR includes work for both CLIENT-8337 and CLIENT-8338. This PR implements lazy signaling where a signaling connection is only made when necessary, i.e. when registering and making a call. It also changes the `Device` states.

### Before review
* [ ] Manually sanity tested running locally
* [ ] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,16 +17,8 @@ enum.
 
 ```ts
 export enum EventName {
-  // Device and Connection errors.
   Error = 'error',
-
-  // Connection events.
-  Cancel = 'cancel',
-  Connect = 'connect',
-  Disconnect = 'disconnect',
   Incoming = 'incoming',
-
-  // Device Registration events.
   Unregistered = 'unregistered',
   Registering = 'registering',
   Registered = 'registered',
@@ -48,22 +40,19 @@ async Device.register(): Promise<void>;
 async Device.connect(options?: Device.ConnectOptions): Promise<Call>;
 ```
 
-- `Device.updateOptions` may be called when there are no connections maintained by
+- `Device.updateOptions` may be called when there are no calls maintained by
 the `Device`. Updating options that change the `Device`'s connection to the
 Twilio backend will cause a re-registration if the `Device` is registered when
 calling `Device.updateOptions`.
 
-- `Device.register` returns a `Promise<void>` such that when the `Device` has
-received a signal from the Twilio backend that the signaling service has
-successfully connected and the `Device` is registered, the returned
-`Promise<void>` will resolve.
+- `Device.register` returns a Promise that resolves when the `Device` has
+successfully connected to the Twilio backend and the `Device` is registered.
 
-- `Device.connect` returns a `Promise<Call>` such that when the `Device` has
-received a signal from the Twilio backend that the signaling service has
-successfully connected and a call has been made, the returned `Promise<Call>`
-will resolve.
+- `Device.connect` returns a Promise that resolves with a `Call` when the
+`Device` has successfully connected to the Twilio backend and a call has been
+made.
 
-Listening for incoming calls:
+#### Listening for incoming calls:
 ```ts
 const token = '...';
 const options = { edge: 'ashburn', ... };
@@ -73,7 +62,7 @@ device.on(Device.EventName.Incoming, call => { /* use `call` here */ });
 await device.register(token, customRegisterOptions);
 ```
 
-Making an outgoing call:
+#### Making an outgoing call:
 ```ts
 const token = '...';
 const options = { edge: 'ashburn', ... };
@@ -82,7 +71,7 @@ const device = new Device(token, options);
 const call = await device.call({ To: 'foobar' });
 ```
 
-Using a device for both incoming and outgoing calls:
+#### Using a device for both incoming and outgoing calls:
 ```ts
 const token = '...';
 const options = { edge: 'ashburn', ... };
@@ -99,7 +88,11 @@ New Features
 
 ### Device Options
 
-The SDK now allows `Device` options to be changed after initial set up.
+The SDK now allows `Device` options to be changed after initial set up using
+`async Device.updateOptions(options?: Device.Options)`. It will return a Promise
+that resolves when the options have been successfully updated within the
+`Device`. If the `Device` is registered before calling `Device.updateOptions`,
+the `Device` will re-register before resolving the returned Promise.
 
 Example usage:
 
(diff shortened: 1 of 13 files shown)


CHANGELOG.md
```diff
@@ -10,72 +10,96 @@ The formula used to calculate the mean-opinion score (MOS) has
...
-using `Device.register(token: string, options?: Device.Options)`.
+```ts
+export enum EventName {
+  // Device and Connection errors.
+  Error = 'error',
+
+  // Connection events.
+  Cancel = 'cancel',
+  Connect = 'connect',
+  Disconnect = 'disconnect',
```
ryan-rowland, line 26: Let's kill cancel, connect and disconnect. These are all emitted from the Call, and are Call-related, not Device.

CHANGELOG.md
```diff
@@ -10,72 +10,96 @@ The formula used to calculate the mean-opinion score (MOS) has
...
+
+async Device.updateOptions(options: Device.Options): Promise<void>;
+async Device.register(): Promise<void>;
+async Device.connect(options?: Device.ConnectOptions): Promise<Call>;
 ```
 
-If `Device` is constructed without any parameters, then it will be instantiated
-with default options. Then, to connect the `Device` to the Twilio servers, a
-call to `Device.register` must be performed.
+- `Device.updateOptions` may be called when there are no connections maintained by
```
ryan-rowland, line 51: We need to start calling Connection -> Call

CHANGELOG.md
```diff
@@ -10,72 +10,96 @@ The formula used to calculate the mean-opinion score (MOS) has
...
+the `Device`. Updating options that change the `Device`'s connection to the
+Twilio backend will cause a re-registration if the `Device` is registered when
+calling `Device.updateOptions`.
+
+- `Device.register` returns a `Promise<void>` such that when the `Device` has
```
ryan-rowland, line 56: "`Device.register` returns a Promise that resolves when..."

CHANGELOG.md
```diff
@@ -10,72 +10,96 @@ The formula used to calculate the mean-opinion score (MOS) has
...
+received a signal from the Twilio backend that the signaling service has
+successfully connected and the `Device` is registered, the returned
+`Promise<void>` will resolve.
 
-`Device.register` returns a `Promise<this>` such that when the `Device` has
+- `Device.connect` returns a `Promise<Call>` such that when the `Device` has
```
ryan-rowland, line 61: Same format suggestion as above

CHANGELOG.md
```diff
@@ -10,72 +10,96 @@ The formula used to calculate the mean-opinion score (MOS) has
...
 received a signal from the Twilio backend that the signaling service has
-successfully connected, the returned `Promise<this>` will resolve with the
-ready to use device.
+successfully connected and a call has been made, the returned `Promise<Call>`
+will resolve.
 
-`Device.setOptions` may be called when there are no connections maintained by
-the `Device`.
+Listening for incoming calls:
```
ryan-rowland, line 66: Denote these as examples, maybe a small header

CHANGELOG.md
```diff
@@ -10,72 +10,96 @@ The formula used to calculate the mean-opinion score (MOS) has
...
 ------------
 
 ### Device Options
 
-The SDK now allows `Device` options to be changed after initial set up. The
-passable `Device` options have changed significantly; options that cause a
-change in the connection to the Twilio backend (such as signaling edge, insights
-options, etc.) are now separated and consolidated within a new interface
-`Device.RegisterOptions`.
+The SDK now allows `Device` options to be changed after initial set up.
```
ryan-rowland, line 102: "..after initial setup using `Device.updateOptions(options: Device.Options)`

Does this return a promise or is this synchronous?

lib/twilio/device.ts
```diff
@@ -491,7 +507,17 @@ class Device extends EventEmitter {
...
-    this.setOptions(options);
+    ([
+      [Device.EventName.Unregistered, Device.State.Unregistered],
+      [Device.EventName.Registering, Device.State.Registering],
+      [Device.EventName.Registered, Device.State.Registered],
+    ] as const).forEach(([event, state]) => {
+      this.on(event, () => {
+        this._state = state;
+      });
+    });
```
ryan-rowland, line 518: A couple of questions:
1. Do we ever unbind these? It doesn't look like we can, given the way this is constructed. This could cause a memory leak.
2. Why are we waiting for the event to be emitted? We should always set the state _before_ emitting the event. Otherwise, the application may get the event before the state has been changed.

lib/twilio/device.ts
```diff
@@ -547,16 +569,16 @@ class Device extends EventEmitter {
...
     if (this._audio) {
       this._audio._unbind();
     }
 
-    if (this._stream) {
-      this._destroyStream();
-    }
+    this._destroyStream();
+    this._destroyPublisher();
+    this._destroyAudioHelper();
```
ryan-rowland, line 581: Do we call any of these methods elsewhere, or what's the reason for moving these all out into their own methods?

lib/twilio/device.ts
```diff
@@ -587,99 +612,80 @@ class Device extends EventEmitter {
...
 
-    if (!hasChunderURIsChanged && this._stream) {
-      this._stream.setToken(this._token);
-    } else {
-      await this._setupStream();
-    }
+    this._shouldReRegister = false;
 
-    return this;
+    this._mediaPresence.audio = false;
```
ryan-rowland, line 674: Can we kill `this._mediaPresence` and just pass true/false to _sendPresence?

lib/twilio/device.ts
ryan-rowland: I think the best approach here is to have a method to set states. Something like:
```
_setState(state: Device.State): void {
  // Don't emit an event if we're already in a state
  if (this._state === state) { return; }
 
  this._state = state;
  this.emit(state);
}
```

Then we never set state or emit independently, we do it by calling this method.

ryan-rowland (changes requested):

lib/twilio/device.ts
mhuynh5757: The stream and publisher ones are called within their respective `setup` functions to tear down existing ones.

PR #275, opened 2021-04-21
[CLIENT-8394] Add "destroyed" device state.

mhuynh5757:
**Contributing to Twilio**

### JIRA link(s):

- [CLIENT-8394](https://issues.corp.twilio.com/browse/CLIENT-8394)

### Description

This PR adds a new `Device` state "destroyed", which is set after `Device.destroy` is called.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,7 +44,7 @@ const interval = setInterval(async () => {
 The Device states have changed. The states were: `[Ready, Busy, Offline]`. These
 have been changed to more accurately and clearly represent the states of the
 Device. There are two changes to Device state:
-1. The states themselves have changed to `[Registered, Registering, Unregistered]`. This
+1. The states themselves have changed to `[Registered, Registering, Unregistered, Destroyed]`. This
 removes the idea of "Busy" from the state, as technically the Device can have an active
 Call whether it is registered or not, depending on the use case. The Device will always
 starty as `Unregistered`. In this state, it can still make outbound Calls. Once `Device.register()`
@@ -52,6 +52,11 @@ has been called, this state will change to `Registering` and finally `Registered
 `Device.unregister()` is called the state will revert to `Unregistered`. If the signaling
 connection is lost, the state will transition to `Registering` or `Unregistered' depending
 on whether or not the connection can be re-established.
+
+The `destroyed` state represents a `Device` that has been "destroyed" by calling
+`Device.destroy`. The device should be considered unusable at this point and a
+new one should be constructed for further use.
+
 2. The busy state has been moved to a boolean, `Device.isBusy`. This is a very basic
 shortcut for the logic `return !!device.activeConnection`.
 
@@ -62,6 +67,7 @@ enum and represent the new Device states:
 
 ```ts
 export enum EventName {
+  Destroyed = 'destroyed',
   Error = 'error',
   Incoming = 'incoming',
   Unregistered = 'unregistered',
@@ -74,6 +80,9 @@ Note that `unregistered`, `registering`, and `registered` have replaced
 `offline` and `ready`. Although frequently used to represent connected or disconnected,
 `ready` and `offline` actually were meant to represent `registered` and `unregistered`,
 which was quite ambiguous and a primary reason for the change.
+
+When the device is destroyed using `Device.destroy`, a `"destroyed"` event will
+be emitted.
 
 ### Device usage changes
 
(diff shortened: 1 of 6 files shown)


lib/twilio/device.ts
```diff
@@ -616,6 +625,8 @@ class Device extends EventEmitter {
    * Register the `Device` to the Twilio backend, allowing it to receive calls.
    */
   async register(): Promise<void> {
+    this._throwIfDestroyed();
```
ryan-rowland, line 628: Looks like this is covered in the condition below

lib/twilio/device.ts
```diff
@@ -583,12 +588,16 @@ class Device extends EventEmitter {
...
     }
+
+    this._setState(Device.State.Destroyed);
   }
 
   /**
    * Disconnect all {@link Call}s.
    */
   disconnectAll(): void {
+    this._throwIfDestroyed();
```
ryan-rowland, line 599: Is this safe? What happens if destroy is called first, while there are ongoing calls? Any chance we can orphan a call?

lib/twilio/device.ts
```diff
@@ -661,6 +672,8 @@ class Device extends EventEmitter {
    * calls.
    */
   async unregister(): Promise<void> {
+    this._throwIfDestroyed();
```
ryan-rowland, line 675: Handled below?

lib/twilio/device.ts
```diff
@@ -563,6 +566,8 @@ class Device extends EventEmitter {
    * Destroy the {@link Device}, freeing references to be garbage collected.
    */
   destroy(): void {
+    this._throwIfDestroyed();
```
ryan-rowland, line 569: We should probably just return here without an error.

lib/twilio/preflight/preflight.ts
```diff
@@ -412,7 +416,9 @@ export class PreflightTest extends EventEmitter {
 
     this._call.once('disconnect', () => {
       this._device.once(Device.EventName.Unregistered, () => this._onUnregistered());
-      this._device.destroy();
+      if (this._device.state !== Device.State.Destroyed) {
+        this._device.destroy();
+      }
```
ryan-rowland, line 421: If we return rather than throw on this condition in destroy() we can revert all the changes in this file.

lib/twilio/device.ts
ryan-rowland: Instead of throwing, I think we should close any open calls and return. If no calls are open we just return.

tests/integration/preflight.ts
```diff
@@ -47,21 +47,23 @@ describe('Preflight Test', function() {
...
   const destroyDevices = () => {
-    if (callerDevice) {
+    if (callerDevice && callerDevice.state !== Device.State.Destroyed) {
       callerDevice.disconnectAll();
       callerDevice.destroy();
+      callerDevice = null;
     }
 
-    if (receiverDevice) {
+    if (receiverDevice && receiverDevice.state !== Device.State.Destroyed) {
```
ryan-rowland, line 56: We can revert these conditions if we return silently from disconnectAll() and destroy()

ryan-rowland (changes requested):

CHANGELOG.md
```diff
@@ -88,15 +97,18 @@ new Device(token: string, options?: Device.Options): Device;
 /**
  * Promise resolves when the Device has successfully registered.
  * Replaces Device.registerPresence()
+ * Can reject if the Device is unusable, i.e. "destroyed".
```
ryan-rowland, line 100: Is this a reject? Thought we were throwing.

lib/twilio/device.ts
```diff
@@ -773,6 +782,12 @@ class Device extends EventEmitter {
    * @param token
    */
   updateToken(token: string) {
+    if (this.state === Device.State.Destroyed) {
+      throw new InvalidStateError(
+        `Attempt to "updateToken" when device is in state "${this.state}".`,
+      );
+    }
```
ryan-rowland, line 789: Why aren't we using the throw function? We can pass in the method name for the InvalidStateError message, eg `this._throwIfDestroyed('updateToken')`

ryan-rowland (changes requested):

lib/twilio/device.ts
ryan-rowland: Never mind :)

CHANGELOG.md
mhuynh5757: Throwing in an async function causes the internal Promise to reject.

lib/twilio/device.ts
mhuynh5757: I was on the fence about this because the error message changes slightly between checks, i.e. in `register` it says that we want the `Device` to be in an "unregistered" state whereas something like `updateToken` doesn't check for an exact state, only that it's not "destroyed".

PR #308, opened 2023-02-03
Removing unreleased changes

charliesantos:
If you look at the commit history since the last release (1.14), there are changes that weren't release. This PR is undoing those changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,31 +1,3 @@
-1.15.0 (In progress)
-====================
-
-New Features
-------------
-
-- ### Twilio Regional Support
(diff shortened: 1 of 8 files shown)


mhuynh5757 (approved):
I'm wondering why we never released these changes.

mhuynh5757:
I see the tests are going kablooey, it's actually the exact same issues I helped @PikaJoyce solve with RTC Diagnostics SDK. We can discuss offline over Slack.

charliesantos:
> I'm wondering why we never released these changes.

It seems they were released in 2.x

PR #309, opened 2023-02-03
VDI Support

charliesantos:
See https://code.hq.twilio.com/client/voice-sdk-api/pull/69 for spec

Tests will come at a later PR


.circleci/images/docker-compose.yml
```diff
@@ -28,9 +28,9 @@ services:
       - "../../:/opt/app"
       - /var/run/docker.sock:/var/run/docker.sock
       - /opt/app/node_modules
-  integrationTests: # runs integration tets. Expects that sources are mounted.
+  integrationTests: # runs integration tests. Expects that sources are mounted.
```
mhuynh5757, line 31: 😆

.gitignore
```diff
@@ -2,7 +2,6 @@
 .idea
 .nyc_output/
 config.yaml
-package-lock.json
```
mhuynh5757, line 5: 👍

lib/twilio/audiohelper.ts
```diff
@@ -165,9 +167,12 @@ class AudioHelper extends EventEmitter {
     this._getUserMedia = getUserMedia;
     this._mediaDevices = options.mediaDevices || defaultMediaDevices;
     this._onActiveInputChanged = onActiveInputChanged;
+    this._enumerateDevices = typeof options.enumerateDevices === 'function'
+      ? options.enumerateDevices
+      : this._mediaDevices && this._mediaDevices.enumerateDevices;
 
     const isAudioContextSupported: boolean = !!(options.AudioContext || options.audioContext);
-    const isEnumerationSupported: boolean = !!(this._mediaDevices && this._mediaDevices.enumerateDevices);
+    const isEnumerationSupported: boolean = !!(this._enumerateDevices);
```
mhuynh5757, line 175: Nitpick: preference for `Boolean(this._enumerateDevices)`

lib/twilio/audiohelper.ts
```diff
@@ -499,11 +504,11 @@ class AudioHelper extends EventEmitter {
    * Update the available input and output devices
    */
   private _updateAvailableDevices = (): Promise<void> => {
-    if (!this._mediaDevices) {
+    if (!this._enumerateDevices) {
```
mhuynh5757, line 507: I see that the other checks use both `!this._mediaDevices || !this._enumerateDevices`, any reason this one is just the latter instead of both?

lib/twilio/log.ts
```diff
@@ -85,7 +90,7 @@ class Log {
    * Set a default log level to disable all logging below the given level
    */
   setDefaultLevel(level: LogLevelModule.LogLevelDesc): void {
-    this._log.setDefaultLevel(level);
+    this._log.setDefaultLevel && this._log.setDefaultLevel(level);
```
mhuynh5757, line 93: Should we warn if `this._log.setDefaultLevel` is not available? In the case where this `&&` guard fails, nothing happens and it silently breaks.

lib/twilio/rtc/peerconnection.js
```diff
@@ -467,15 +478,20 @@ PeerConnection.prototype._updateAudioOutputs = function updateAudioOutputs() {
 };
 
 PeerConnection.prototype._createAudio = function createAudio(arr) {
-  return new Audio(arr);
+  const audio =  new Audio(arr);
```
mhuynh5757, line 481:
```suggestion
  const audio = new Audio(arr);
```

lib/twilio/rtc/peerconnection.js
```diff
@@ -467,15 +478,20 @@ PeerConnection.prototype._updateAudioOutputs = function updateAudioOutputs() {
...
-  this._mediaStreamSource.connect(dest);
+  let dest = null;
+  if (this._mediaStreamSource) {
+    dest = this._audioContext.createMediaStreamDestination();
+    this._mediaStreamSource.connect(dest);
+  }
 
   const audio = this._createAudio();
-  setAudioSource(audio, dest.stream);
+  setAudioSource(audio, dest && dest.stream ? dest.stream : this.pcStream);
```
mhuynh5757, line 494: Same effect, but we could use nullish coalescing if you think it makes the code less verbose.
```ts
dest?.stream ?? this.pcStream
```

lib/twilio/audiohelper.ts
charliesantos: Good observation. I think it should check for both. Will update.

lib/twilio/log.ts
charliesantos: Good idea.

lib/twilio/rtc/rtcpc.js
```diff
@@ -4,13 +4,15 @@ const Log = require('../log').default;
...
+function RTCPC(options) {
   if (typeof window === 'undefined') {
     this.log.info('No RTCPeerConnection implementation available. The window object was not found.');
     return;
   }
 
-  if (util.isLegacyEdge()) {
+  if (options && options.RTCPeerConnection) {
+    this.RTCPeerConnection = options.RTCPeerConnection;
+  } else if (util.isLegacyEdge()) {
```
mhuynh5757, line 15: Should we have the `else if` here? What happens if an override is passed, but somehow we detect legacy Edge?

charliesantos: It doesn't matter if we detect Edge or any browser really if there's an override. The customer overriding RTCPeerConnection will handle all the APIs themselves.

charliesantos: Oh and to answer the question "What happens if an override is passed, but somehow we detect legacy Edge?", we will still use the override no matter what.

mhuynh5757 (approved):

lib/twilio/rtc/peerconnection.js
charliesantos: Just tried and I'm getting a syntax error. I think it's because our build tools are out of date. I'm reverting to old syntax for now.

PR #313, opened 2023-02-28
Citrix VDI Support

charliesantos:
Merge PR for Citrix VDI Support


lib/twilio/device.ts
```diff
@@ -1122,7 +1124,6 @@ class Device extends EventEmitter {
     const region = getRegionShortcode(payload.region);
     this._edge = regionToEdge[region as Region] || payload.region;
     this._region = region || payload.region;
-    this._publisher.setHost(createEventGatewayURI(payload.home));
```
mhuynh5757, line 1125: We got rid of this because we had dangling home region work right?

charliesantos: Yes. Check this merge commit https://github.com/twilio/twilio-client.js/pull/308/files

mhuynh5757 (approved):

